PR #892, opened 2014-07-08
Initial support for BananaPi

CReimer:
These three commits provide support for BananaPi to uboot, linux-armv7 and linux-armv7-rc

There is an upstream known bug left, which more or less breaks ethernet support. It works only with 100Mbps and even then it's not as stable as my USB ethernet card.

And I'm still confused by all these device tables. It's booting and it's stable. Nevertheless I'm sure that some minor features are still not working (e.g status LED)


synthead:
I'm disappointed that this pull request was denied because of subjective opinions about this SoC module.  This is Arch Linux ARM, and this is an ARM board.  What's the deal?  Isn't Arch Linux ARM supposed to be an open project?

CReimer:
I sold my BananaPi because of this and replaced it with a Cubieboard 2.
As far as I know BananPi support was added to upstream uboot a few weeks ago.

PR #1059, opened 2015-01-08
Added phantomjs fixes for ARM

d235j:
This has been tested working on armv7h. It needs testing on the other platforms.
The fixes are ported from the qt4 PKGBUILD (phantomjs bundles Qt). It appears that -fno-strict-volatile-bitfields is no longer needed.

### DIFF
--- a/community/phantomjs/PKGBUILD
+++ b/community/phantomjs/PKGBUILD
@@ -1,0 +1,46 @@
+# $Id$
+# Maintainer: Felix Yan <felixonmars@gmail.com>
+# Contributor: grimsock <lord.grimsock at gmail dot com>
+# Contributor: Dieter Plaetinck <dieter@plaetinck.be>
+# Contributor: Vladimir Chizhov <jagoterr@gmail.com>
+# Contributor: Henry Tang <henryykt@gmail.com>
+
+# ALARM: Kevin Mihelich <kevin@archlinuxarm.org>
+#  - Just say NO to x86 optimizations, and no neon
+#  - added -fno-strict-volatile-bitfields to CXXFLAGS to fix ARM bug
+
+
+pkgname=phantomjs
+pkgver=1.9.8
+pkgrel=1
+pkgdesc="Headless WebKit with JavaScript API"
+url="http://www.phantomjs.org/"
+license=('BSD' 'LGPL' 'MIT')
+arch=('i686' 'x86_64')
+depends=('gstreamer0.10-base' 'fontconfig' 'freetype2' 'gcc-libs')
+makedepends=('git')
+source=("git+https://github.com/ariya/${pkgname}.git#tag=$pkgver")
+
+build() {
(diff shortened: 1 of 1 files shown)


kmihelich:
No distcc should not be needed here; qt4 is a very special case with that.

d235j:
PhantomJS bundles a patched-up copy of Qt 4.8.4 that *must* be used.

WarheadsSE:
Since this bundles it's own qt4, would it still needed?

kmihelich:
No distcc should not be needed here; qt4 is a *very* special case with that.

d235j:
@kmihelich: can you describe the reasons for no distcc, and why it would not be needed here? I'd test this with distcc enabled but that's not going to help he here.

d235j:
Updated to enable distcc (I'm informed that it's disabled in qt4 because qt4 propagates build settings to programs built against it — would have been nice if this was in the notes for qt4 since this would make things much clearer here).

community/phantomjs/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+# Contributor: grimsock <lord.grimsock at gmail dot com>
+# Contributor: Dieter Plaetinck <dieter@plaetinck.be>
+# Contributor: Vladimir Chizhov <jagoterr@gmail.com>
+# Contributor: Henry Tang <henryykt@gmail.com>
+
+# ALARM: Kevin Mihelich <kevin@archlinuxarm.org>
+# (From qt4-4.8.4-18 PKGBUILD)
+#  - Just say NO to x86 optimizations, and no neon
+#  - added -fno-strict-volatile-bitfields to CXXFLAGS to fix ARM bug
+#  - not necessary to disable distcc because other software won't be built against phantomjs qt4
```
kmihelich: it's really not necessary to say something isn't necessary, or that the notes are from another package.

d235j: Should this instead say that phantomjs bundles its own Qt4 (which is not entirely obvious at first glance)?

kmihelich: No.  Remove both.

community/phantomjs/PKGBUILD
```diff
@@ -0,0 +1,48 @@
...
+
+
+pkgname=phantomjs
+pkgver=1.9.8
+pkgrel=1
+pkgdesc="Headless WebKit with JavaScript API"
+url="http://www.phantomjs.org/"
+license=('BSD' 'LGPL' 'MIT')
+arch=('i686' 'x86_64' 'armv7h')
```
kmihelich: Do not need to specify our architectures, we build with -A.

community/phantomjs/PKGBUILD
d235j: Is it appropriate to leave this line with your name? What is the correct format for this line in general?

Issue #1075, opened 2015-01-30
 in core/linux-raspberrypi/config edit config


pums974:
Remove CONFIG_REGULATOR - we don't have a regulator driver 

Basically apply this upstream commit :
https://github.com/raspberrypi/linux/commit/8605f6d9cfbdb8c83c49382c02f8ba6475529e2c

which solve this bug (digi+)
http://forum.kodi.tv/showthread.php?tid=192380&pid=1836662#pid1836662

PR #1076, opened 2015-01-30
Update config

pums974:
issue  #1075
Remove CONFIG_REGULATOR - we don't have a regulator driver

Basically apply this upstream commit :
raspberrypi/linux@8605f6d

which solve this bug (digi+)
http://forum.kodi.tv/showthread.php?tid=192380&pid=1836662#pid1836662


kmihelich:
Manually editing the config is unacceptable.  That is why it is IN CAPS AT THE BEGINNING OF THE FILE.

The PKGBUILD also needs to be updated to reflect the changed config.

PR #1077, opened 2015-01-30
Remove CONFIG_REGULATOR - we don't have a regulator driver

pums974:
Basically apply this upstream commit :
raspberrypi/linux@8605f6d

which solve this bug (digi+)
http://forum.kodi.tv/showthread.php?tid=192380&pid=1836662#pid1836662

(the right way this time, sorry for the inconvenience)


kmihelich:
Why is the pkgrel not incremented?

pums974:
Because i'm an idiot

pums974:
Thank you for the merge

PR #1078, opened 2015-01-31
[odroid-c1] Add udev rules to fix /dev/am* permissions

Holzhaus:
This fixes permissions for the /dev/am* devices on ODROID-C1. I hope this is the right package to put them into (if not, please tell me which packe this belongs to and I'll open a new PR).

I tested the udev rules on my C1 and they work fine.

Related forum thread: [http://archlinuxarm.org/forum/viewtopic.php?f=63&t=8348](http://archlinuxarm.org/forum/viewtopic.php?f=63&t=8348)


pepedog:
I don't think this belongs in a kernel package. What about one of the alarm/Odroid-c1- packages?

Holzhaus:
What do you think about [alarm/aml-libs-c1](https://github.com/archlinuxarm/PKGBUILDs/tree/master/alarm/aml-libs-c1)?

kmihelich:
This would make more sense in aml-libs-c1, since you need those drivers to use the devices anyway.

Holzhaus:
Ok, this PR now refers to the alarm/aml-libs-c1 package.

PR #1083, opened 2015-02-04
Enable IPsec in kernel

bl0ndynek:
As you mention on forum: 
http://archlinuxarm.org/forum/viewtopic.php?f=31&t=8339
I create pull request to enable IPsec in kernel by default.


kmihelich:
*Why* does this need to be built-in? Do the modules not work?

pepedog:
I am guessing you manually edited this file, as bfq patches are not available.

bl0ndynek:
I enable it in kernel.

@kmihelich could you tell me what module should I load to enable it? it could be enough.

kmihelich:
Each CONFIG option has documentation that is easily found, and which specifies what its module name will be.  This documentation is not me.

PR #1084, opened 2015-02-05
Enable again CONFIG_DVB_AS102 for linux-raspberrypi

EliaCereda:
This option was enabled in 3.12.x kernels and I think it should still be enabled in 3.18.x.

I use this driver myself and it'd be very useful to have it enabled by default again because I wouldn't have to manually recompile the kernel every time.


kmihelich:
Editing the config by hand is unacceptable, and this doesn't update the PKGBUILD.

EliaCereda:
Why close a pull request like this? Both things you mentioned can be fixed easily. What would be the acceptable way to request something like this?

Here's a diff with the changes made by the linux build system when enabling that option in `make nconfig`. The `CONFIG_DVB_AS102_FE` had to be enabled too, so I made a commit that does it:

```
--- config	2015-02-05 17:51:29.999958189 +0100
+++ src/linux/.config	2015-02-05 18:16:56.458975834 +0100
@@ -1,6 +1,6 @@
 #
 # Automatically generated file; DO NOT EDIT.
-# Linux/arm 3.18.4-1 Kernel Configuration
+# Linux/arm 3.18.5 Kernel Configuration
 #
 CONFIG_ARM=y
 CONFIG_SYS_SUPPORTS_APM_EMULATION=y
@@ -2690,7 +2687,7 @@
 CONFIG_SMS_USB_DRV=m
 CONFIG_DVB_B2C2_FLEXCOP_USB=m
 # CONFIG_DVB_B2C2_FLEXCOP_USB_DEBUG is not set
-# CONFIG_DVB_AS102 is not set
+CONFIG_DVB_AS102=m
 
 #
 # Webcam, TV (analog/digital) USB devices
@@ -2896,7 +2893,7 @@
 CONFIG_DVB_RTL2830=m
 CONFIG_DVB_RTL2832=m
 CONFIG_DVB_SI2168=m
-# CONFIG_DVB_AS102_FE is not set
+CONFIG_DVB_AS102_FE=m
 
 #
 # DVB-C (cable) frontends
```

I've also updated the PKGBUILD with the new checksum. Is there anything else that needs to be done to make this pull request mergeable?

WarheadsSE:
This being missing points us to the fact that it was not done with make (?)config. You can see how this being missing is a key point.
```
-# Linux/arm 3.18.4-1 Kernel Configuration
+# Linux/arm 3.18.5 Kernel Configuration
```
Otherwise, looking at the current diff, the pkgrel and md5sums are now updated.

kmihelich:
Included in 3.18.5-2.

PR #1097, opened 2015-02-14
[alarm/kodi-c1] Fix wrong version number

Holzhaus:
Package: `kodi-c1`
Architecture: `armv7h`
Device: ODROID-C1

The PKGBUILD states that it's version 14.0, but in facts it’s 15.0-ALPHA1 (that's at least what Kodi says in the "System Information" panel and on the splash screen).
Thus, this PR changes the `pkgver` to `15.0alpha1` and resets `pkgrel` to 1.
As the Kodi codename has not yet been announced, I changed the `_codename` to `I*******`, which is consistent with the codename displayed on the [splash screen](https://github.com/mdrjr/xbmc/blob/k_c1/media/Splash.png).

Splash Screen


alarm/kodi-c1/PKGBUILD
```diff
@@ -7,6 +7,7 @@
 # Contributor: BlackIkeEagle < ike DOT devolder AT gmail DOT com >
 # Contributor: Bartłomiej Piotrowski <bpiotrowski@archlinux.org>
 # Contributor: Maxime Gauduin <alucryd@gmail.com>
+# Contributor: Jan Holthuis <holthuis.jan@googlemail.com>
```
kmihelich: You're a contributor for changing the version?

Holzhaus: I thought you do this everytime you change a package. If you don't want that, I'll leave it out. I'm not that vain ;-)

kmihelich: You add your name when you have contributed something to the package.

kmihelich:
Assuming this package reaches version 15.0, 15.0alpha1 < 15.0.  This version is unusable.

Holzhaus:
OK, what should I change this to? 15.0?

Holzhaus:
Don't want to annoy you, but I changed the version to `15.0` (without `alpha1`).

PR #1105, opened 2015-02-18
/alarm/odroid-c1-libgl: Add odroid-c1-libgl-fb-headers

Apokalypzx:
Add the Mali framebuffer headers for the odroid-c1. Change odroid-c1-libgl-headers to odroid-c1-libgl-x11-headers.


kmihelich:
There are more packaging mistakes than I have time to go into, and this would create a conflict with the same headers provided by mesa, leading to an uninstallable package.

Holzhaus:
@kmihelich Do you think It would be sane to put these into something like `/opt/mali/include` instead?

PR #1112, opened 2015-02-23
core/linux-raspberrypi: Add support for tcp_lp module

drizzt:
it can be used in trasmission to avoid saturate the bandwidth.
see http://www.pps.univ-paris-diderot.fr/~jch/software/bittorrent/tcp-congestion-control.html


WarheadsSE:
pkgrel not bumped.
checksum not updated.

You made this change with menuconfig/nconfig/etc?

PR #1113, opened 2015-02-24
WiringX - A modular GPIO library for several platforms

crazystick:
WiringX provides a modular approach to GPIO access, based on WiringPi,
but offering a consistent API across multiple platforms. Of the
platforms supported by Arch Linux Arm, it currently works on Raspberry
Pi 1 / 2 and Hummingboard (Cubox-i)


kmihelich:
This can easily be combined into one PKGBUILD with split packages for better maintainability.  Makedepend on python and python2, then add depends in package()'s respectively.  Depends/makedepends on packages in base-devel should not be included, those are assumed to be present, and unused PKGBUILD variables can be deleted.

crazystick:
Thanks for the advice, hopefully that looks better now, I haven't created a split package before.

kmihelich:
Makedepends have no effect inside package() functions, they are global only. Also good to define pkgbase on split packages to avoid ambiguity (use "wiringx-git").  Makedepends on 'python-setuptools' and 'python2-setuptools' will pull in python and python2, so those are redundant.

PR #1121, opened 2015-03-02
core/linux-raspberrypi: Update Device Tree config

michaellass:
* fix wrong name of overlay for w1-gpio using external pullup
* update w1-gpio(-pullup) parameters according to /boot/overlays/README


kmihelich:
Why is the checksum in the PKGBUILD not updated?

PR #1133, opened 2015-03-11
Add alarm/weston-rpi

plfiorini:
This package builds weston with the Raspberry Pi compositor on armv7h too.

As a matter of fact the weston package from community only support
the Raspberry Pi armv6h models, hence this package is needed for
proper Raspberry Pi 2 support.

### DIFF
--- a/alarm/weston-rpi/PKGBUILD
+++ b/alarm/weston-rpi/PKGBUILD
@@ -1,0 +1,47 @@
+# $Id$
+# Maintainer: Sébastien Luttringer
+# Contributor: Joel Teichroeb <joel@teichroeb.net>
+
+# ALARM: Kevin Mihelich <kevin@archlinuxarm.org>
+#  - export flags for rpi
+
+# ALARM: Pier Luigi Fiorini <pierluigi.fiorini@gmail.com>
+#  - build for armv7h
+
+buildarch=4
+
+pkgname=weston-rpi
+pkgver=1.7.0
+pkgrel=1
+pkgdesc='Reference implementation of a Wayland compositor (Raspberry Pi 2)'
+arch=('armv7h')
+url='http://wayland.freedesktop.org'
+license=('MIT')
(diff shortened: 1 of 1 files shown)


kmihelich:
Given that the v6 repo is only used on the pi, we can continue to support the changes for v6 in the community package.  This package only makes sense to build for v7.

alarm/weston-rpi/PKGBUILD
```diff
@@ -0,0 +1,46 @@
...
+pkgname=weston-rpi
+pkgver=1.7.0
+pkgrel=1
+pkgdesc='Reference implementation of a Wayland compositor'
+arch=('armv6h' 'armv7h')
+url='http://wayland.freedesktop.org'
+license=('MIT')
+depends=('libxkbcommon' 'libinput' 'libunwind' 'poppler-glib' 'mtdev' 'libxcursor' 'glu' 'pango' 'colord')
+makedepends=('raspberrypi-firmware')
+replaces=('weston')
```
kmihelich: This should not replace weston.  There are, in fact, other ARM devices in existence than the pi.

plfiorini:
@kmihelich having two different packages for weston on raspberry seems confusing to me, i would preferer if there was only this one.

alarm/weston-rpi/PKGBUILD
plfiorini: but this is meant for the raspberry pi only and users should not install it on other devices.
besides the replaces line will make this replace weston with a working one on an existing rpi2 installation.

kmihelich:
In the context of this PR, my last note states what needs to happen for acceptance.  If you have something else in mind, then change your PR to reflect that.

alarm/weston-rpi/PKGBUILD
kmihelich: And it will replace any installed weston package on any other device with an rpi-specific one.  Please review our list of supported devices to confirm for yourself that there are, in fact, other devices aside from the pi, and we support many of them.
http://archlinuxarm.org/platforms

plfiorini:
PR updated.

Issue #963, opened 2014-09-09
docker: 1.2.0 not recognizing btrfs filesystem


djmaze:
Package: docker
Architecture: armv7
Device: ODROID U3
 docker daemon: 1.2.0 fa7b24f-dirty; execdriver: native; graphdriver: btrfs

What it isn't doing: 

The docker daemon won't start anymore after the upgrade. Instead, the journal tells me:

    prerequisites for driver not satisfied (wrong filesystem?)

What it should be doing: 

The daemon should start successfully.

Steps to reproduce:

* Run docker 1.0.0 with a btrfs filesystem mounted at `/var/lib/docker`. (Need to add `-s btrfs` to the `ExecStart` line in _/usr/lib/systemd/system/docker.service_ beforehand.) 
* Upgrade docker to 1.2.0. 
* Re-add the `-s btrfs` flag to the _docker.service_ file.  
* `sudo systemctl daemon-reload && sudo systemctl restart docker`  

Thoughts:

I recognize the [btrfs detection patch has been removed during the upgrade to 1.2.0](https://github.com/archlinuxarm/PKGBUILDs/commit/f5a9cd5c82ac810836561fd9d05a580f36a140d4#diff-2). Maybe this is still needed?

asssaf:
Is there a workaround?

djmaze:
Did not yet have an opportunity to look at this again. Will get a new device in the coming days where I will try to get this working, hopefully making some progress.

asssaf:
I've been testing this in QEMU so far.
If you'd like I test your work on my Pi.

WarheadsSE:
This does really need updated too...

djmaze:
Ok, so I just tried re-adding the `0002-Fix-btrfs-detection-on-ARM.patch` to the latest 1.4.1 PKGBUILD. I tested it on my new ODROID-C1. Btrfs is working fast and flawlessly as a storage driver. Moreover, it does not exhibit slow container deletion times, including errors in syslog, as I am experiencing with the default devicemapper driver. 

From my point of my view, btrfs seems to be the better choice. The compatibility patch really should be added back. I can make a PR if you wish.

kmihelich:
Yes, please submit a PR if this has been fully tested.

djmaze:
[...]

PR #1135, opened 2015-03-14
core/linux-odroid-c1: Include btrfs in C1 kernel

djmaze:
This changes the kernel for the ODROID-C1 to include btrfs in the kernel as opposed to a module. This allows using brtfs as a root filesystem. (`/boot` must reside on a separate ext*/fat/whatever partition, as AFAIK Uboot does not support booting from btrfs yet.)

I am using this to benefit from btrfs advantages compared to ext4: Easier "partition" management, backups and rollbacks through subvolumes, multi-device support. Also, a better experience with Docker (see https://github.com/archlinuxarm/PKGBUILDs/issues/963#issuecomment-72756639). So you just need one small ext4 partition for the static boot files and one big btrfs partition for everything else.


kmihelich:
Was the config edited by hand?

djmaze:
Yes.. I was not sure which was the best way on how to do this. I should better do it again with `make menuconfig` in order to get the additional options in the config, I suppose?

kmihelich:
Yes, you should.  Hence why the very top of the config says: Automatically generated file; DO NOT EDIT.

djmaze:
Ok, sorry. Will update this.

djmaze:
Ok, updated. I also enabled btrfs debugging support, as it is enabled for ext4 as well.

Also confirmed it is working.

kmihelich:
Is debugging required?

djmaze:
From my experiences, there are no problems with btrfs in the latest kernels anymore. So we might as well disable debugging.

djmaze:
Disabled debugging support now.

PR #1136, opened 2015-03-18
fixes for trimslice an others

tmn505:
Common fixes and enhancements
-CONFIG_CFG80211_INTERNAL_REGDB=y
Removes warnings about empty db.txt see explanation => https://git.kernel.org/cgit/linux/kernel/git/stable/linux-stable.git/tree/net/wireless/db.txt
+CONFIG_DVB_AS102=m
enabled in 3.17 an 3.18 lost in transition to 3.19

Trimslice fixes
+CONFIG_RTC_DRV_EM3027=y
the module is not inserted automatically during boot
+asoc-tlv320aic23-input-signals-routing-warning.patch
remover warnings about null routing of input signals

All compiled and tested on TrimSlice


WarheadsSE:
Patch name is out of convention, showing the order in which it is applied. Should be 0006*

kmihelich:
It's not named correctly because it is a copy and paste with no headers or attribution, and has not been added into my patch set using git.

tmn505:
Well I'm new to git, and this was my usual manner of working with source, should I replace it with patch command?

kmihelich:
I've added the changes into the packages myself.

PR #1137, opened 2015-03-19
server: bb.org: migrate rcn-ee.net to rcn-ee.com

RobertCNelson:
I have a new dedicated box running rcn-ee.com, to prevent any possible
downtime as i move rcn-ee.net over, switch to the .com now.

btw: https is also available, if you want to eventually enable that too.

Signed-off-by: Robert Nelson <robertcnelson@gmail.com>


kmihelich:
Thanks for the heads up, and especially for the work you do.

WarheadsSE:
+5

PR #1142, opened 2015-03-20
kodi-rbp: fix bug in PKGBUILD that prevents kodi.service from shutdown and restart

graysky2:
* add polkit as a dep to allow kodi.service to shutdown and restart
* fix the install path to make the included rules active


WarheadsSE:
This change is far more sane.

graysky2:
Agreed >:)

graysky2:
If you decide to merge it, you consider the other kodi packages offered as well.  Your policy on PRs clearly states one per package.

kmihelich:
As the changes are *identical*, how about you consider the other kodi package this clearly applies to as well.

WarheadsSE:
Also, would this particular change be needed for kodi-imx?

graysky2:
I haven't inspected your other kodi packages, but I believe that if they are all based on the same parent skeleton, this should be ported to them as well.  I will leave that decision to you guys since I only have a RPi for testing and only use [alarm]/kodi-rbp myself.

kmihelich:
So, to be clear, you are refusing to acknowledge and comply with the request to add these changes to the git version?

graysky2:
No, I first want to you or someone to verify that my PR is a valid solution.  I will gladly port it to the kodi-rbp-git package since I can test that on my hardware.  I am uncomfortable applying to other kodi packages that won't run on my hardware and that I can therefore not test.

kmihelich:
Then once again, you are submitting a PR you have not personally validated?

graysky2:
kmilhelich - How did you come to that conclusion????

kmihelich:
Because you just said, "... I first want to you or someone to verify that my PR is a valid solution."

graysky2:
Yes, I would think that you would want that level of peer-review before committing a change, no?  In any case, if you verify you like this PR, I will port into the git package.  I don't want to do double work so please let me know that you approve.

moonman:
Would it make more sense to add polkit to optdepends?

graysky2:
From my perspective, if you're going to ship a service files that needs polkit to work with 100 % of the features (ie shutdown/reboot) the package itself should depend on it.

moonman:
I know this was done in the past. Polkit is not required to run kodi, but adds extra functionality. Though I understand where you are coming from.

graysky2:
How ever you guys want to handle it is fine with me.  Most Arch users will read the optdeps statements from pacman upon installation.  The main thing to change is the PKGBUILD installing the polkit rules in a location where they will be active if polkit is present on the system.

moonman:
fixed in https://github.com/archlinuxarm/PKGBUILDs/commit/68bb89c2a515f9d70cec0d87ed334b2e87874b30

NicoHood:
I recently tried the latest kodi-rbp and I do not have shutdown buttons for the kodi service. I have not installed any DE. I checked all opt dependencies but there was no suitable hint.

timjklein36:
Also recently installed latest kodi-rbp, however, no shutdown options are available in the menu in kodi. I am using the standalone with no window manager. Should I install all opt deps?

graysky2:
@timjklein36 - I believe both upower and polkit should be installed for this to work.  Not 100% sure about upower.

PR #1144, opened 2015-03-22
Add armv7h target (Raspberry Pi 2)

larsch:


larsch:
Disregard, not compatible with RPi2

PR #1145, opened 2015-03-22
libbcm2835 1.42, RPi2 support

larsch:


kmihelich:
This will not build for v7 in our build system.  Read the README again.

Where did that patch come from, and why is it needed?

larsch:
buildarch updated.

I created the patch. The inline assembly in the upstream is not accepted by current gcc assembler due to missing newlines.

kmihelich:
You need to clean up the patch and get rid of the extraneous text.

larsch:
Fair. Done.

PR #1146, opened 2015-03-24
Enable ioquake3-rbp on Raspberry Pi 2

larsch:


kmihelich:
Have you test built this (in a clean chroot)?  This package has not been touched in a long time.

larsch:
I had, but not in chroot. It think I have fixed it now (problem was chmod of a symlink/readonly file). At least it Also cleaned up the source/git repo handling. Wish I had discovered devtools-alarm a bit earlier :-)

PR #1171, opened 2015-04-24
Hifiberry passthrough support

baranator:
Hi,

i recently bought a Raspberry Pi 2 and a Hifiberry digi+. I wanted to connect the raspberry pis hdmi to my tv and give the sound via spdif/toslink-passthrough to my av-receiver. But I encountered the problem that the receiver only got stereo from the raspberry allthough the played files had a 5.1 audio-track.
I noticed that the "digital passthrough device" in Settings/System/Audio was grayed out and permanently set to "HDMI". After some search i found [this commit](https://github.com/OpenELEC/OpenELEC.tv/blob/f6d2a67e07f3380a6960809c9071ee3bbbf5d672/projects/RPi/patches/kodi/kodi-001-isengard-rpb-backports.patch) in OpenElec that claimed to solve this problem by enforcing the hifiberry digi to be recognized as a digital output device. And indeed after patching and rebuilding, passthrough-output-selection and audio-passthrough finally worked as expected.

As the hifiberry is a popular extension for the raspberry pi and its (for now partly broken) key feature is to extend its audio capabilities i think this hotfix is quite useful.

### DIFF
--- a/alarm/kodi-rbp/PKGBUILD
+++ b/alarm/kodi-rbp/PKGBUILD
@@ -25,15 +25,17 @@ makedepends=('hicolor-icon-theme' 'fribidi' 'lzo2' 'smbclient' 'libtiff' 'libva'
              'udisks' 'upower' 'git' 'autoconf' 'java-environment')
 source=("https://github.com/xbmc/xbmc/archive/$pkgver-$_codename.tar.gz"
         'kodi.service'
-        'polkit.rules')
+        'polkit.rules'
+	'xbmc-14.2-Helix.patch')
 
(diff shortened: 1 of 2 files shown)


kmihelich:
Has there been a patch submitted for the kernel upstream?  That appears to be the correct solution.

moonman:
pkgrel needs to be bumped too.

baranator:
i just took the patch from openelec and adapted it to match this versions source. I don't even know which kernel-module this affects (i suppose snd-soc-wm8804 ?)

PR #1200, opened 2015-06-08
core/linux-parallella to 3.14.12

sctincman:
This uses the newer 3.14 kernel from Adapteva which includes the new Epiphany device driver (which is needed for the newer eSDK).

The kernel config was adapted from their defconfig, and modified to resemble the previous Arch config


kmihelich:
This config strips out too much support for device drivers and features that people expect to be present.  The defconfig is a reference, not a final draft.

kmihelich:
The commented header install lines can likely be uncommented now with the improved config.  Also, squash your commits into one.

PR #1201, opened 2015-06-08
alarm/parallella-firmware: add updated headless 7010/7020 variants

sctincman:
This adds two new packages for the [updated] headless 7010/7020 parallella bitstreams.

This is based on the 2015.1 release, which only provides the headless variants. HDMI variants can be added once those bitstreams are hosted somewhere.

This also removes the older alarm/parallella-firmware package as it is out of date.

These packages provide/conflict "parallella-firmware" and the hope is a user can install whichever matches their board by selecting {HDMI,headless}-{7010,7020}


kmihelich:
I'd rather see this as a single PKGBUILD with split packages, since that will decrease future administration overhead, and these are nearly identical PKGBUILDs using a common commit base.

sctincman:
Very true, just pushed that change.

PR #1204, opened 2015-06-16
dkms-7601: Changed upstream source to https://github.com/porjo/mt7601

ddaygold:
I had been making fixes to my own copy of the source, but this new
upstream does it better.

I've started with the AUR mt7601-git PKGBUILD, and added dkms.

I also fixed dkms to actually work.

### DIFF
--- a/alarm/dkms-mt7601/PKGBUILD
+++ b/alarm/dkms-mt7601/PKGBUILD
@@ -1,41 +1,30 @@
-pkgname=dkms-mt7601
-pkgver=v3.0.0.4
-pkgrel=2
-pkgdesc="Driver for Ralink MT7601 chipset wireless adaptors"
-arch=('any')
-url="http://www.ralinktech.com"
-license=('GPL')
-depends=('dkms' 'linux-headers')
-conflicts=()
-install=${pkgname}.install
-options=(!strip)
-_pkgname="mt7601"
-source=("https://ddaygold.github.io/DPO_MT7601U_LinuxSTA_3.0.0.4_20150203.pfc.tar.bz2"
-        "dkms.conf")
+# Maintainer of AUR mt7601-git package: Cyker Way < cykerway at gmail dot com >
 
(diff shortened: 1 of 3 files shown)


WarheadsSE:
It does look like this will likely be pulling into mainline at that moment. My only concern is making this a severe note that there is no compatibility with before 3.19

ddaygold:
Sorry, I'm having a hard time parsing your comment.

Do you mean this PR will be merged, or that there's a mainline alternative?

And yes, this package broke in 3.19 kernels b/c of a naming change to a struct. Should I change 
```
depends=('linux>=3.0')
```
to 
```
depends=('linux>3.19')
```

I apologize for my lack of knowledge, this is my second pull request!

WarheadsSE:
Yup. Please update that exact part :)

ddaygold:
Done.

I don't know if you get notifications about commits, but you clearly do about messages. Hopefully I haven't double spammed you!

kmihelich:
Squash your commits.

ddaygold:
Done.

I didn't realize that you could squash commits once you published them to GitHub. I thought the hard and fast rule was 'don't change shared history'. Would'a done it sooner.

kmihelich:
Have you made sure this builds?

ddaygold:
Tested builds on x86_64 and a beagle bone black. It works on both.

As a bonus, while testing I also fixed the .install file to correctly remove the module! 

Should've caught that the first time.

kmihelich:
There is at least one error that will cause the build to fail.  I suggest you build using a clean chroot.

ddaygold:
I checked again, but I don't see the problem. To be clear, it *does* build, install and function right now. Tested on x86_64 and a BBB.

What am I doing wrong?

kmihelich:
Not building with a clean chroot. This package *will not* build in our build system as it is now.

ddaygold:
I read the manual. 
multilib-build works now. Am I missing other problems?

PR #1206, opened 2015-06-19
Add default lang to plex-media-server

dz0ny:
Refused to start without language defined in env


WarheadsSE:
This sounds like a failed system environment setup, not a packaging issue. Is your locale.conf correct, and properly used locale-gen?

kmihelich:
Configuring your locale is *your* responsibility, not some package's.

dz0ny:
This was fresh install (os and package) start.sh distributed in package also does this. Also locale.conf is not reflected in plain systemd env...

PR #1207, opened 2015-06-19
Add Banana Pi

seidler2547:
Banana Pi is already supported upstream


kmihelich:
Companies that spam our site and social media with their unoriginal crap will not be supported here.

PR #1208, opened 2015-06-25
add i.MX6/7 kernel

l3iggs:
This is a kernel that should work with anything containing Freescale's i.MX6 or i.MX7 processors.

Tested on my TBS2910:
http://www.tbsdtv.com/launch/tbs-2910-matrix-arm-mini-pc.html


kmihelich:
No. Use linux-armv7.

l3iggs:
Have you seen linux-armv7 working on a TBS2910?

When I try to use linux-armv7 on my hardware, it hangs silently after u-boot successfully loads it.
This kernel works.

WarheadsSE:
Find the difference, do a PR to fix that config, etc.

Also, don't check in .SRCINFO, especially with lots of worthless whitespace.

PR #1219, opened 2015-07-23
kodi-rbp upgrade to 15.0 Isengard

beam:


davidbeauchamp:
Manually bumped.

beam:
And https://github.com/beam/archlinuxarm-pkgbuilds/commit/168779bc4082407787be5cbc81788e9cfd4d9066 ?

beam:
And https://github.com/beam/archlinuxarm-pkgbuilds/commit/168779bc4082407787be5cbc81788e9cfd4d9066 ? 
Kodi 15.0 require libcec 3.0, we have only 2.2

moonman:
Beam is right, without the patch CEC will be disabled. Though since nothing else uses libcec-rpi I will bump it to v3 and will need to rebuild both kodi-rbp-git and kodi-rbp.

beam:
It's taken from kodi-rbp-git, so only kodi-rbp is needed to rebuild
https://github.com/archlinuxarm/PKGBUILDs/blob/master/alarm/kodi-rbp-git/PKGBUILD#L38

moonman:
I added the patch to revert libcec3 because at the time kodi-rbp was using
v2.2. There is no need in reverting to 2.2 now

beam:
can you upgrade libcec-imx6? i have to use same patch in https://github.com/archlinuxarm/PKGBUILDs/pull/1220

moonman:
libcec3 doesn't compile, probably due to GCC or cmake being too new. Let's take this discussion to the forum.

davidbeauchamp:
libcec 3.0.1, libcec-rpi 3.0.1 and kodi-rbp 15.0-2 have been built and pushed out.

PR #1237, opened 2015-08-25
Updated Cubox-i U-Boot to mainline 2015.07.

Frans-Willem:
Boot script was changed (to keep deviations from mainline minimal), but support for uEnv.txt was added in case you need something specific.
Tested with linux-armv7 on a Cubox-i4pro.


Frans-Willem:
WarheadsSE suggested changing ext4load to load for it to be filesystem agnostic. I've got this working and compiling, so I'll submit a new pull-request for that. Sorry for the trouble!

PR #1238, opened 2015-08-25
Updated Cubox-i U-Boot to mainline 2015.07.

Frans-Willem:
Boot script was changed (to keep deviations from mainline minimal), but support for uEnv.txt was added in case you need something specific.
Tested with linux-armv7 on a Cubox-i4pro.

(Newer version of https://github.com/archlinuxarm/PKGBUILDs/pull/1237 with ext4load changed to load per WarheadSE's instructions)


moonman:
please squish the commits.
Also the commit message should specify what package is getting updated i.e.
alarm/uboot-cubox-i to 2015.07-1
Thanks

Frans-Willem:
Removed all the commits that had appeared due to merging with the main branch, and changed the commit message as instructed.
Anything else that I should change?
And would anyone be willing to point me to getting my own branch up to date with the archlinux one, without adding extra commits :-/ ?

moonman:
This will have to be reverted. After some research I found at least 2 regressions:
1. On reboot kernel gets stuck at some point. I didn't bother to check with serial.
2. Hummingboard 2 is not supported in the mainline (yet)

PR #1241, opened 2015-08-31
rethindb

MarcMcIntosh:
added rethinkdb-git package


kmihelich:
This is not the AUR.

MarcMcIntosh:
so where's it going? (if anywhere)..... it's only been built on/for arm

PR #1243, opened 2015-09-02
PhantomJS PKGBUILD modified to add support for ARM

barnabycolby:
Taken from the Arch community repository, the PKGBUILD has been updated to pass an additional '--jobs 2' parameter to the PhantomJS build script, this allows the build to succeed on devices with a lower amount of memory, typical of ARM devices. The build can, depending on the system take as long as several days, so building this package for ARM users will likely save them a great deal of time.

When trying to get the package to build on my raspberry pi's, running out of memory resulted in strange compilation errors that caused me some confusion, automating this will save ALARM users some headaches.

### DIFF
--- a/community/phantomjs/PKGBUILD
+++ b/community/phantomjs/PKGBUILD
@@ -1,0 +1,56 @@
+# $Id$
+# Maintainer: Felix Yan <felixonmars@archlinux.org>
+# Contributor: grimsock <lord.grimsock at gmail dot com>
+# Contributor: Dieter Plaetinck <dieter@plaetinck.be>
+# Contributor: Vladimir Chizhov <jagoterr@gmail.com>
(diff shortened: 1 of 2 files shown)


kmihelich:
This package is already in the repositories.

barnabycolby:
Could you link me to it please? I'm having trouble finding it.

moonman:
```pacman -Ss phantomjs
community/phantomjs 2.0.0-4
    Headless WebKit with JavaScript API
```

barnabycolby:
Ah, I got confused because i thought that the only packages built for ARM were those with PKGBUILD scripts in this repository. The archlinuxarm.org website states that this is not the case:
'core, extra, community: Packages in these folders (unless for good reason) are the packages found in the upstream repos by the same names. These represent packages that needed to be changed from the upstream version, not the only ones we build.'

PR #1254, opened 2015-09-13
Adds opencascade that builds for armv7h

tirithen:
I have modified and added a PKGBUILD (and unmodified env.sh, opencascade.conf, opencascade.sh) for opencascade (required by freecad). After some trial and error it does build and install on my armv7h (ODROID-XU4), added short description of the patching I've done in the top of PKGBUILD.

I have also added and modified freecad by just adding armv7h to the arch array inside PKGBUILD and tested compile, install and start the freecad program. It seem to be running fine. :)

This is the first time I do this so it might be a good idea to look through this, I have only done the simplest things to get it building and installing.

Would be great to be able to provide opencascade (takes several hours to build on the ODROID-XU4) and freecad as prebuilt packages.

I only have this arm so I could not try any of the other arm architectures.

### DIFF
--- a/community/freecad/PKGBUILD
+++ b/community/freecad/PKGBUILD
@@ -1,0 +1,77 @@
+# Maintainer: Florian Pritz <bluewind@xinu.at>
+# Contributor: Jonas Heinrich <onny@project-insanity.org>
+# Contributor: Jonas Heinrich <onny@project-insanity.org>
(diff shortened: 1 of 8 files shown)


tirithen:
Please tell me if there are something that I can do to ease and speed up the process of getting this package into the repo. :)

PR #1269, opened 2015-09-26
xf86-video-armsoc-odroid: Fix build failure with xserver version 1.17

gripped:
There is a commit upstream at http://cgit.freedesktop.org/xorg/driver/xf86-video-armsoc/commit/src/drmmode_display.c?id=25f588e2c4addb74ef0f730485878610073b6470
which deals with the issue.
We only want this one commit as others cause issues with fullscreen especially on the Odroid XU3/4

Without this patch xf86-video-armsoc-odroid does not build successfully.
With the patch it does build. And is tested to work.

It does not work well on the XU3/4 and never has. Frequently, switching programs from fullscreen to windowed results in a black screen with only the mouse cursor showing.
Should this pull request be deemed acceptable to be merged I will create another soon with a armsoc package specific to the XU3/4 which fixes this.


kmihelich:
Merge the changes into the other patch that encompasses the same goal.

gripped:
Done

PR #1270, opened 2015-09-28
mpv: Enable egl video output

gripped:
mpv supports egl video output but it has to be specifically enabled by configure.
Since version 0.10.0 the 'x11' video output module has been removed.
On an odroid at least the 'xv' video output module does not work so without the egl module no video output seems possible.

The actual command line switch to enable the egl video output is not that easy to find and I considered adding a message in the .install file. I can add it if you like.
For the record it is
`--vo=opengl:backend=x11egl:es`

I've tested this with the (Mali) egl/gles libraries removed from my device, and it does still build. In fact the egl output still works (via mesa, with poor quality). So this pull release should have no impact on devices without egl/gles.

### DIFF
--- a/community/mpv/PKGBUILD
+++ b/community/mpv/PKGBUILD
@@ -9,7 +9,7 @@
 pkgname=mpv
 epoch=1
 pkgver=0.11.0
-pkgrel=1
+pkgrel=1.1
 pkgdesc='Video player based on MPlayer/mplayer2'
 arch=('i686' 'x86_64')
 license=('GPL')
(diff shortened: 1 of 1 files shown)


gripped:
And sorry I should have added mpv: to the title of the pull release.

kmihelich:
Has this been tested to build on at least v5, v6, and v7 with a clean chroot?

gripped:
I don't have a v5 or a v6 device. 
I have tested building in a clean chroot with three separate makepkg.conf files with the CARCH, CHOST and C(XX)FLAGS as they would be on each type of device.

In each case it built successfully.

PR #1274, opened 2015-10-11
jxcore

MarcMcIntosh:
adding an architecture specific version of the jxcore PKGBUILD in the aur


kmihelich:
Why does this need to be in our repos?

MarcMcIntosh:
it doesn't need to be there grand scheme of the universe, but the ability to compile and run optimised serverside javascript on arm based devices has it's uses. (for more info search for asm.js on wikipedia)

Although, one issue i was having with the installation was a conflict with some of the node header files provider by both node and jxcore.... ended up uninstalling node and resinstalling it using a node version managment system (n).

moonman:
I suppose the question was more about why does it have to be in ALARM repo, rather than stay in AUR and more importantly who is going to maintain it here? You can ask the maintainer in AUR to add your changes. You can also become a co-maintainer in AUR of the package(new AUR4 feature).

MarcMcIntosh:
It's probally more uefull on the alarm repo, or arm devices.... at least for what I'm doing and my current work flow. write a client, and a server app, webpack them in to two js files move to an arm server so  compilling the server.js in to asm.js has it's benifits.

I am the aur maintainer, so it's almost like hitting two birds with one stone.

PR #1277, opened 2015-10-12
patch blender cycle optimisations, so it can be build on non-x86 systems

Popolon:
A test missed in cycle SIMD optimisations, that used by default x86 optimisations, this patch show at the same time how to do when x86 and armv7 are used at the same time, if someone add neon optimizations one day...

### DIFF
--- a/community/blender/PKGBUILD
+++ b/community/blender/PKGBUILD
@@ -1,0 +1,109 @@
+# $Id$
+# Contributor: John Sowiak <john@archlinux.org>
+# Contributor: tobias <tobias@archlinux.org>
+# Maintainer: Sven-Hendrik Haase <sh@lutzhaase.com>
+
+# ALARM: Popolon <popolon@popolon.org>
+#  - patch to fix building on non-x86 platforms (this doesn't add neon optimized code)
+
+# Sometimes blender.org takes some time to release patch releases and because Arch users
+# are impatient, we sometimes need to build from git directly.
+# Update because I get so many queries on this:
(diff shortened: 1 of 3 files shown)


kmihelich:
Does this successfully build using a clean chroot on all four supported architectures?

WarheadsSE:
And has it been upstreamed already?

Popolon:
I compiled it in ALARM ARMv7h environment. it has been upstreamed with a simplified form patch, I suppose it will be here for next Blender version. The modification doesn't impact x86 build if x86/x86_64 is still used it continue to use x86 SIMD optimizations, I can simplify it as Blender team done if you want.

Popolon:
The patch on the trunk:
https://git.blender.org/gitweb/gitweb.cgi/blender.git/commit/ea7ce7fcf417c77434b67dead12002b20b3bf59f

WarheadsSE:
If it will be here next blender version, and does not provide immediate
value, then this should not be merged, because it is going against the arch
way of simplicity

Popolon:
It allows users to use blender on archlinux ARM now (there is currently no ALARM version of Blender at all). Blender 2.76 was released on October 11, 2015. Previous version was on july 8 2015, (2.75a), and 2.75 on july 1st, so perhaps a patched 2.76a will be released soon ?? previous version was 2.74, released in march 31 without 2.74a release, so perhaps we will have to wait 3 months.

I made a precompiled package anyway that can be downloaded here:
http://popolon.org/depots/ArchlinuxARM/blender-17:2.76-1.0001-armv7h.pkg.tar.xz

PR #1279, opened 2015-10-15
Add support for Broadcom's VC4 hardware to mesa

greysAcademicCode:


WarheadsSE:
My only immediate concern is that this would add this to all 4 architectures, where the VC4 is only applicable for v6/v7

greysAcademicCode:
Should I put an if statement around it that checks the architecture?

WarheadsSE:
Perhaps something simpler, like a conditional filling a variable that fills the end of the line

greysAcademicCode:
You're thinking of a line like `[ $CARCH == "armv7h" -o $CARCH == "armv6h" ] && VC4=',vc4'`?

greysAcademicCode:
making new PR for this

PR #1283, opened 2015-10-18
Add support for Broadcom's VC4 hardware to mesa

greysAcademicCode:


kmihelich:
Why should this be built? This is for the open-source kernel driver, which is not what is used here.  This is also still under development.

And stop opening new PRs for the same thing.  GitHub documents how to use pull requests if you are confused.

joaodriessen:
Mesa 11.1 is the development release.
The kernel driver for the vc4 has been included in the current release of the the raspberry pi kernel for Arch Linux Arm.

greysAcademicCode:
I've just resolved all the merge conflicts with this, should be ready to merge again now.

pisto:
Do I understand correctly that this enables the already upstreamed vc4 work, which however does not offer the same features of the raspbian distro as of yet?

PR #1293, opened 2015-11-07
kodi-rbp-git to 16.0a4-Jarvis

phedoreanu:


alarm/kodi-rbp-git/PKGBUILD
```diff
@@ -44,10 +44,10 @@ install:' -i tools/EventClients/Makefile.in
 }
 
 build() {
-  cd "${srcdir}/xbmc-${_commit}"
+  cd "${srcdir}/xbmc-${_tag}"
 
   # Bootstrapping
-  MAKEFLAGS=-j1 ./bootstrap
+  MAKEFLAGS=-j5 ./bootstrap
```
WarheadsSE: Do not do this. Hard-coding that bootstrap to more cores than other users might have is bad practice. -j1 was set because problems with mutli-process in the past. At worst, this should be removed.

phedoreanu: Closed this PR and opened #1294.

WarheadsSE: Please just update the code and squash commits in the future. No need for
repetitious PR :)

phedoreanu: I wanted to squash, but the commits were in the reverse order and I don't know how to rebase and change the parent, so I did a push -f. Sorry for repetition.

PR #1297, opened 2015-11-11
extra/mesa enable freedreno

andersson:
Signed-off-by: Bjorn Andersson <bjorn.andersson@sonymobile.com>


andersson:
Tested on my Xperia Z2 Tablet.

PR #1302, opened 2015-11-17
Update community/docker to 1.9.0

djmaze:
Works for me. 

Note I am using an experimental 4.2 kernel on ODROID-XU4 with the (non-default) overlayfs driver. This should probably be tested by someone with an older kernel as well.

### DIFF
--- a/community/docker/PKGBUILD
+++ b/community/docker/PKGBUILD
@@ -2,8 +2,8 @@
 # Maintainer: Sébastien "Seblu" Luttringer
 
 pkgname=docker
-pkgver=1.7.1
-pkgrel=2
+pkgver=1.9.0
+pkgrel=1
 epoch=1
 pkgdesc='Pack, ship and run any application as a lightweight container'
 arch=('x86_64')


djmaze:
Well, that was much better! Thanks.

luxas:
Have you come across this? #1305

djmaze:
@kmihelich, unfortunately it seems the 1.9.0 version is not working reliably for most people (see the [Docker bug report](https://github.com/docker/docker/issues/18125)). Something seems to be broken when Docker is built using Go 1.5 on ARM (regardless of the actual Docker version).

I don't know how the ALARM build infrastructure works. Is it possible to build the Docker package using a Go 1.4 version? Or just replace the binary package in the repository with one that was built using Go 1.4? 
 
Alternatively, can the package be downgraded in the repository, until the issue has been resolved? Although I am not sure that will work so people can downgrade the package on their machines afterwards?

On an related note: I recognize the Github issue tracker for this repository has been disabled. Is there another location for people to report bugs? (The website still references the issues at Github).

alexellis:
@djmaze +1 @kmihelich Docker 1.9.x does not work at all on ARM this this commit moving forward from Docker 1.7.1 and Go 1.4.x. Are the package maintainers aware of this?

kmihelich:
Yes, I've seen these comments and the bug report, and this is on the Docker developers to fix. When they fix it, this package can be updated to incorporate those changes.

alexellis:
@kmihelich OK, so while waiting for Docker ARM to work with Go 1.5 - I've been told building a Docker binary from 1.9.1 works fine with Go 1.4.3. I was mid-way through the build when I got this error:

```
/usr/bin/ld: cannot find -ldevmapper
/usr/bin/ld: cannot find -ldevmapper
collect2: error: ld returned 1 exit status
/usr/local/go/pkg/tool/linux_arm/5l: running gcc failed: unsuccessful exit status 0x100
```

I have all dependencies including device-mapper. Do you know how I can satisfy this dependency? I am reading as cannot find device-mapper headers - but this is definitely installed.

alexellis:
@kmihelich @djmaze I have been able to build Docker on ALARM using the official method of building Docker on Docker. I have uploaded the working build scripts to my github repository if you want to take a look?

https://github.com/alexellis/docker-arm/tree/master/images/docker-arm

Ideally alarm should not use the hack scripts, but a version of docker itself to build a binary. 

The main file - Dockerfile.armhf which builds Docker 1.9.1 installs a binary version of Go 1.4.3 which is possibly why this works and the Arch version doesn't as discussed previously.

kmihelich:
The PKGBUILD here is what is used in upstream Arch Linux.

Ideally, Docker should QA their software against current versions of the tools they depend on.  Go 1.5 was released August 2015.  This was before 1.9 was even a release candidate.

PR #1316, opened 2015-12-06
Add support for Acer Chromebook CB5-311

RaumZeit:
The patches included in this branch add support for the Acer Chromebook CB5-311, and possibly other Nyan Chromebooks. They provide (i) the 3.10.18 ChromeOS kernel from googles git repo, (ii) the proprietary NVIDIA Tegra K1 video drivers, and (iii) alsa UCM config to support the audio devices when using ChromeOS instead of mainline kernel.

I am happy to change whatever is necessary to make this a proper merge, it it isn't already.

Best,

Ronny


kmihelich:
Squash your commits into one per package, and use the existing chromebook kernel packages as a template for source retrieval and installation.

RaumZeit:
Done... The kernel image resembles the one used for peach chromebook. It only contains minor changes in both, PKGBUILD and .config, necessary to deal with NVIDIA Tegra, and v3.10 specific config parameters. Kernel sources themselves are now downloaded as tarball of a specific commit instead of pulled from HEAD of Chromium git repo.

RaumZeit:
I'll withdraw this pull request for now, since it seems that there are too many issues still open for a successful merge that meets the specifications.
It would be nice, If you could guide me a little regarding the mrvl firmware files that need to be taken from ChromeOS filesystem. For linux-peach, the corresponding binaries are hosted on www.archlinuxarm.org. Would this be possible for the nyan specific files, too?

Also, I am wondering what the right partition layout for chromebooks is. My Samsung Chromebook (linux-peach) has 2 partitions, one named 'Kernel' the other 'Root'. On install of the kernel, the install script determines the device name of the root partition and the kernel is flashed to the first partition of this device. So I guess the specification is one kernel partition (1st partition), one rootfs (somewhere else on the same device). Am I assuming that right?
The internal eMMC storage of my Acer CB5-311 was repartitioned, so that the kernel partition is p6, rootfs is p7. Would ALARM allow for such configurations as well?

I hope, the packages can still be merged some time, because I think the remaining issues are just minor ones.

Hoping for some suggestions,

Best
Ronny

PR #1317, opened 2015-12-08
Adding support for Acer Chromebook CB5-311 (2nd attempt)

RaumZeit:
Hi,
since my first attempt was flawed by not following the Arch Linux ARM guidelines, I restructured my packages for a second try.

**Changes to my previous pull request (#1316)**
The PKGBUILD for the ChromeOS kernel is almost identical to that of core/linux-peach. Required firmware blobs are downloaded from google, and archlinuxarm.org. The kernel patches included in the package allow for proper wifi support, and ensure that the kernel can be build with GCC5. Default flash device for the kernel follows the same strategy as linux-peach, i.e. if device partition number is 2, kernel goes to partition 1 of the same device, otherwise flash target is the partition containing the rootfs. However, I added another target to keep this package compatible for those people who already installed Arch Linux ARM with my modified install script (https://github.com/RaumZeit/LinuxOnAcerCB5-311). Here, the default target partition is number 6 if rootfs is located on partition 7. I hope this minor addition is ok.

The gpu-nvidia-tegra-k1 PKGBUILD does not contain instructions for the TegraK1 accelerated  libjpeg.so anymore. As far as I know, there is no safe way to package this library.

**Installation procedure for Acer CB5-311**
The kernel package was tested using the official Arch Linux ARM installation procedure for Samsung Exynos Chromebook using an USB stick, but I replaced the rootfs tarball with ArchLinuxARM-armv7-latest.tar.gz. After extraction of the rootfs to USB, I chrooted into the installation, installed my linux-nyan kernel package and the following three additional packages to get wifi-menu working:
```
dialog
iw
wpa_supplicant
```
My Acer CB5-311 happily booted from the USB stick, xorg runs fine with fbdev.

### Description of the three packages in this branch:
The patches included in this branch add support for the Acer Chromebook CB5-311, and possibly other Nyan Chromebooks. They provide (i) the 3.10.18 ChromeOS kernel from googles git repo, (ii) the proprietary NVIDIA Tegra K1 video drivers, and (iii) alsa UCM config to support the audio devices when using ChromeOS instead of mainline kernel. The latter is especially useful for scenarios where pulseaudio is used as a layer between application and alsa.

*Please note, that the linux-nyan, and gpu-nvidia-tegra-k1 PKGBUILDs were already available through my github page in late August this year. Hence, the release numbers actually reflect the releases since then, and do not start with 1.*

I hope that the packages are better suited for a merge into archlinuxarm PKGBUILDs repository now. Still, I'd love to contribute and am very much willing change whatever is necessary.

Best,

Ronny


kmihelich:
You need to fix those kernel patches to use the actual commits with the correct author attribution, not just patching whatever you find in and creating a commit.

alarm/gpu-nvidia-tegra-k1/PKGBUILD
```diff
@@ -0,0 +1,194 @@
...
+
+  # copy udev rules
+  mkdir -p "${pkgdir}/usr/lib/udev/rules.d/"
+  install -D -m0644 "${srcdir}"/51-nvrm.rules "${pkgdir}/usr/lib/udev/rules.d/51-nvrm.rules"
+
+}
+
+package_gpu-nvidia-tegra-k1-x11() {
+  pkgdesc="NVIDIA binary OpenGL|ES libraries and X11 drivers for Tegra K1"
+  depends=('gpu-nvidia-tegra-k1="${pkgver}"' 'X-ABI-VIDEODRV_VERSION=19')
```
kmihelich: This is incompatible with the current xorg-server.

RaumZeit:
Ok, I am half way through including the original commits to the kernel. But I am struggling a little while trying to apply commits from mainline kernel that need to be applied to files in different subdirectories. In particular, the chromeos-3.10 kernel makes use of a directory drivers/net/wireless-3.8 that requires changes. But mainline commits correspond to drivers/net/wireless instead. I did not find a way to, e.g. cherry-pick these commits from mainline into chromeos-3.10 without spoiling the commit history. Is there any way to do this properly?
Another problem I face is that one of the patches in my kernel package is actually part of a much larger mainline commit. Here again, I have this wireless vs. wireless-3.8 subdir problem. Moreover, the mainline commit changes not only files in this particular subdirectory, but more than 200 files all over the kernel tree, which need to be dropped before application to chromeos-3.10. So what is a proper way to include this change? Does it suffice to patch the specific file in question and include a reference comment in the commit message?

alarm/gpu-nvidia-tegra-k1/PKGBUILD
RaumZeit: Shall I remove the entire gpu-nvidia-tegra-k1 PKGBUILD, or just the x11 driver? Until the update to xorg-server 1.18 about a month ago, this package was working fine.

RaumZeit:
Sorry for closing this pull request again, but I think it is best, if I start from scratch with just the kernel package. I am sure, this will help in the review process. Also, is it better to discuss possible merges in the Arch Linux ARM forum, or should I make pull requests here on github first?

PR #1319, opened 2015-12-09
add core/linux-nyan

RaumZeit:
This package adds the ChromeOS kernel of branch _chromeos-3.10_ as used for Nyan based Chromebooks. The PKGBUILD is almost identical to that of core/linux-peach. Required firmware blobs are downloaded from google, and archlinuxarm.org. The kernel patches included in the package allow for proper wifi support, and ensure that the kernel can be build with GCC5.2. I branched off the chromeos-3.10 kernel and applied all the included patches on top of this tree (see cherry picking messages in the patches for the corresponding source). Most of them were backported from mainline linux-stable, and only some of them were directly applied by me or can  be found elsewhere on the web, e.g. Kernel Mailing List.
The PKGBUILD specifies a default flash device for the kernel that follows the same strategy as linux-peach, i.e. if device partition number is 2, kernel goes to partition 1 of the same device, otherwise flash target is the partition containing the rootfs. However, I added another target to keep this package compatible for those people who already installed Arch Linux ARM with my modified install script (https://github.com/RaumZeit/LinuxOnAcerCB5-311). Here, the default target partition is number 6 if rootfs is located on partition 7. I hope this minor addition is ok.

**Installation procedure for Acer CB5-311**
The kernel package was tested using the official Arch Linux ARM installation procedure for Samsung Exynos Chromebook using an USB stick, but I replaced the rootfs tarball with ArchLinuxARM-armv7-latest.tar.gz. After extraction of the rootfs to USB, I chrooted into the installation, installed my linux-nyan kernel package and the following three additional packages to get wifi-menu working:

```
dialog
iw
wpa_supplicant
```

My Acer CB5-311 happily booted from the USB stick, xorg runs fine with fbdev.
I also need to say that this kernel (and earlier commits of this branch) are running stable on my Acer CB5-311 since August 2015.

_Please note, that the linux-nyan PKGBUILD was already available through my github account in late August this year. Hence, the release number actually reflects the releases since then, and does not start with 1._


github-login:
Any comment on this 1 year old PR? What it is the reasons it is not accepted ?

aleden:
There is no need to use an old crappy kernel. I am writing this right now on a Acer Chromebook 13 CB5-311 (i.e. nyan-big) running a stock ALARM linux-armv7 kernel:

[    0.000000] Linux version 4.11.8-1-ARCH (builduser@leming) (gcc version 7.1.1 20170516 (GCC) ) #1 SMP Fri Jun 30 01:50:47 UTC 2017

Everything is working great. To get it working, I "chainloaded" u-boot following the directions here:

https://github.com/u-boot/u-boot/blob/v2017.07-rc3/doc/README.chromium#L37

ChristophHaag:
Mainline is not working great.

mwifiex is very unreliable. Try disconnecting and connecting to a wifi a couple of times. It should very quickly produce some bug messages in dmesg and you won't be able to reconnect to any wifi unless you reboot.

cpufreq is completely disabled. I have no idea at what cpu speed it runs but it might just be the highest one.

LPAE is not enabled in archlinux arm mainline kernel. That means you only have 2 gigabyte ram.

Waking up from suspend is buggy. since 4.10 or 4.11 I think. The graphics in X are all messed up unless you change the resolution.

There are really a couple of issues more that I don't remember right now but all in all my experience has been bad enough that I installed the old kernel, X and proprietary nvidia driver from https://www.tbi.univie.ac.at/~ronny/alarm/?C=M;O=A. Turns out it can't suspend at all, but at least the rest of the system works a lot better. Only X feels weird, like somehow the fillrate or so is severely reduced.

aleden:
ok here's what you can do:

```
git clone https://github.com/archlinuxarm/PKGBUILDs.git
cd PKGBUILDs
git checkout b910e4a0e12289dd67a46cd5a086f0330b87bc30
cd core/linux-armv7
makepkg -so
cd src/linux-4.11
make nconfig
# System Type -> [*] Support for the Large Physical Address Extension                                                                                         
cp .config ../../config
cd ../..
makepkg -si --noextract
```

Until the ALARM developers make a LPAE kernel this is all you have to do!

> Mainline is not working great.


mainline is the only way to live, man! shit *actually does* improve over time (I'm talking a looong time, this is fucking linux).

so subscribe to linux-tegra then you can report bugs like I do https://marc.info/?l=linux-tegra&m=149943410200537&w=2 and we can try to make the world a better place.

if it weren't for that ^^^bug then I would be using 4.12 on my chromebook, because it fixes the CPU frequency bug:

> $ [aeden@rota ~]$ sudo cpupower frequency-info
>  analyzing CPU 0:
>   driver: cpufreq-dt
>   CPUs which run at the same hardware frequency: 0 1 2 3
>   CPUs which need to have their frequency coordinated by software: 0 1 2 3
>   maximum transition latency: 300 us
>   hardware limits: 204 MHz - 2.32 GHz
>   available frequency steps:  204 MHz, 306 MHz, 408 MHz, 510 MHz, 612 MHz, 714 MHz, 816 MHz, 918 MHz, 1.02 GHz, 1.12 GHz, 1.22 GHz, 1.33 GHz, 1.43 GHz, 1.53 GHz, 1.63 GHz, 1.73 GHz, 1.84 GHz, 1.94 GHz, 2.01 GHz, 2.12 GHz, 2.22 GHz, 2.32 GHz
>   available cpufreq governors: conservative ondemand userspace powersave performance schedutil
>   current policy: frequency should be within 204 MHz and 2.32 GHz.
>                   The governor "performance" may decide which speed to use
>                   within this range.
>   current CPU frequency: 2.32 GHz (asserted by call to hardware)

and it might even fix your troubles with mwifiex_sdio (i'm sorry you've had them; i haven't).

ChristophHaag:
Actually mainline has gotten worse. I remember 4.5 working rather nicely, but now it's gotten worse. Here is everything: https://archlinuxarm.org/forum/viewtopic.php?f=49&t=9688

tl;dr: The latest attempt at a mainline config with minor patches is in this fork btw: https://github.com/sctincman/PKGBUILDs/tree/nyan-emc-timings/core/linux-armv7-rc
Several of the patches have been upstreamed, not sure if all.

I've asked in the forums about maybe a separate kernel with lpae enabled but apparently, no interest: https://archlinuxarm.org/forum/viewtopic.php?f=23&t=11227

With the 4.11-rc kernels I already got a black screen at boot, with 4.12-rc the same. The release of 4.11 worked normally though. I always thought it wasn't booting at all, but maybe it was just the display not working...

My impression was that nobody is testing nyan-big anymore with suspend having gotten less reliable, wifi getting worse etc. For example afaik https://github.com/torvalds/linux/commit/80373d37bee562060ce4d3e5272b708e8db5d1fc has never been revisited.

aleden:
Ok, I can see that you've worked a lot on this. What a headache.

I recorded a short video that shows how my nyan-big chromebook running under 4.11.8-1-ARCH performs (showing suspend & wifi).

https://drive.google.com/open?id=0B-Z9DeEVbfuNYUZFMTZhN0JWTlk

Not shown in the video but also working:
(1) backlight via echo 256 > /sys/class/backlight/backlight/brightness
(2) battery usage via tlp-stat --battery (however one needs to echo bq24735-charger > /etc/modules-load.d/battery.conf.

kmihelich:
You can continue this conversation on the forum.

PR #1324, opened 2016-01-02
Fix kodi-rbp-git crash on startup

julianxhokaxhiu:
Use CX11 ABI

### DIFF
--- a/alarm/kodi-rbp-git/PKGBUILD
+++ b/alarm/kodi-rbp-git/PKGBUILD
@@ -53,7 +53,7 @@ build() {
   # Configuring Kodi
   export PYTHON_VERSION=2  # external python v2
   # we need to compile for armv6 instead of armv5 to avoid problems compiling assembler code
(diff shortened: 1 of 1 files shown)


julianxhokaxhiu:
@kmihelich can you please merge this?

pepedog:
This won't automatically build and end up in repo, release needs bumping for that.

julianxhokaxhiu:
Who should we bump then?

moonman:
Since the forum is not available ATM, I'll reply here as well. This will not work. 
kodi stable for multiple devices was built manually using gcc5.1. Then kmihelich, once again, manually pushed it to the repos. kodi-rbp-git is a low priority package and seeing how many rebuilds it requires it will not be rebuilt manually every time. That is until the problem with gcc > 5.1 is identified and fixed.

Your proposed fix will not have any effect with gcc 5.3 since gcc 5.3 already defaults to the new abi. It was indroduced for an older gcc 5.1 package which defaulted to the old abi.

julianxhokaxhiu:
Thank you very much for the explanation :)

PR #1325, opened 2016-01-13
[alarm/kodi-rbp-git] Apply CXX11_ABI to fix SIGBUS crash

jellysheep:
This applies commit 6e7c288 to the kodi-rbp-git package to avoid the crash with a SIGBUS error as mentioned here: http://archlinuxarm.org/forum/viewtopic.php?f=15&t=9599.
Without this patch, this package is currently unusable on Armv7h as Kodi immediately crashes on startup.


jellysheep:
Thanks a lot!

moonman:
As I was saying in another PR this will have no effect as other packages (kodi stable) were built with gcc 5.1. Current gcc5.3 already defaults to the new abi. But I guess PRs wont stop untill this is committed and people realise it doesn't help.

PR #1327, opened 2016-01-20
encapsulate dependency on llvm 3.5

wamserma:
As there is no open issue tracker for ALARM I'm abusing a PR.
Thanks for reintroducing ```ghc```. But the dependency on llvm35 leads to a conflict with an installed llvm. I propose to add Tom Hall's [```ghc-llvm```](https://github.com/Thra11/PKGBUILDS) which would allow ```ghc``` to coexist with an up-to-date ```llvm```

see also:
https://thra11.github.io/posts/2015-09-07-GHC-On-Arch-Linux-Arm.html


kmihelich:
No, you will not abuse this as an issue tracker. It will be used what it is designed to be used for.

PR #1329, opened 2016-01-24
Resolved some issues with Jxcore installation and simplified the config arguments 

MarcMcIntosh:
PKGBUILD is now the same as the one in the AUR


WarheadsSE:
Can you please squash this?

Also, I don't think that we can accept a commit message of "changes to be committed"

MarcMcIntosh:
Sure, 
guess the squash by rebase-ing  didn't work. 
I'll try a reset and see how that goes.

Thanks @WarheadsSE

PR #1331, opened 2016-01-24
 Re-wrote PKGBUILD in aur/jxcore.

MarcMcIntosh:
also sorry i may have made wee mistake bit squashing a few commits and endined up making thousands more earlyer on.... I think it's fixed now


kmihelich:
This doesn't need to be in our repo.

MarcMcIntosh:
where would be a good place for it?

PR #1337, opened 2016-02-05
added dub package manager for the D programming language

yannick:
dub should be specifically built for gdc, otherwise it won't be usable


yannick:
since ARM has no DMD the linke flags need to be adjusted for dub, see:
https://github.com/D-Programming-Language/dub/blob/master/build-gdc.sh

further more the current binary package isn't even working as it links against a wrong libphobos.
if you compile it as is it may build but the resulting binary will not work 
(e.g. `dub init -t vibe.d foo` will crash)

PR #1340, opened 2016-02-08
[kodi-rbp] Trigger rebuild to build against new libmicrohttpd

jellysheep:
The updated libmicrohttpd package now provides libmicrohttpd.so.12 instead of libmicrohttpd.so.10:
`/usr/lib/kodi/kodi.bin: error while loading shared libraries: libmicrohttpd.so.10: cannot open shared object file: No such file or directory`
This commit just increases `pkgrel` to trigger a package rebuild.


kmihelich:
Duplicate.

PR #1347, opened 2016-03-09
core/linux-raspberrypi: Support automatically generating initrd.

AlD:
Checks `/boot/config.txt` for an `initramfs <initramfs file>` line and generates `<initramfs file>` using `mkinitcpio`.
Without this, people using initramfs have to remember to regenerate the initramfs manually after a kernel update, because they can otherwise end up with an unbootable system.


AlD:
Can some please tell me why this is being ignored?

At least reject it for some reason, please.

kmihelich:
Not an acceptable generation method.

PR #1348, opened 2016-03-14
Add LFS for PHP on 32-bit

Xenopathic:
On 32-bit architectures, glibc will use 32-bit integers for filesizes, which have a maximum limit of ~2GB. Compiling programs with `-D_FILE_OFFSET_BITS=64` allows a program to use 64-bit integers instead, with no other code changes. However, it seems that this change is incompatible with Apache, and will cause it to hang. This PR brings a new set of packages, `php-lfs-*`, which are compiled with the 64-bit flag and replace the standard `php-*` packages with the appropriate `replaces`, `conflicts` and `provides` options. There is no `php-lfs-apache` for obvious reasons, but all other packages are mirrored in this way.

I haven't performed the clean chroot build test yet, will update with the progress (and whether this flag actually works or not)

### DIFF
--- a/extra/php-lfs/PKGBUILD
+++ b/extra/php-lfs/PKGBUILD
@@ -1,0 +1,369 @@
+# $Id: PKGBUILD 261071 2016-03-06 10:48:44Z pierre $
+# Maintainer: Pierre Schmitz <pierre@archlinux.de>
+# Compiled with large file support for 32-bit ARM platforms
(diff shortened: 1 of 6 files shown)


WarheadsSE:
Why should this be placed into extra? There is no upstream Arch package named this in extra.

kmihelich:
This isn't the AUR.  I see no reason for this to be in our repos, particularly when it's replacing the original php packages.

PR #1365, opened 2016-05-07
 MongoDB: Force mmapv1 to be compiled in

TPXP:
By default, MongoDB 3.2.5 dropped the only storage engine that was available on 32-bits ARM platforms : `mmapv1`. This prevents MongoDB from working properly as seen [on the forums](https://archlinuxarm.org/forum/viewtopic.php?f=15&t=10236).

This commit forces mmapv1 to be included on all ARM platforms, so that MongoDB works properly even straight out of the box. Indeed, on first start, MongoDB will try to initialize the database using the storage engine in the configuration file (`mmapv1`), which was not included by default.

### DIFF
--- a/community/mongodb/PKGBUILD
+++ b/community/mongodb/PKGBUILD
@@ -10,9 +10,12 @@
 #  - patches
 #  - specify mmapv1 as the default storage engine in mongodb.conf
 
+# ALARM: Thomas P. <tpxp@live.fr>
+#  - force mmapv1 to be compiled in the build
+
 pkgname=mongodb
 pkgver=3.2.6
-pkgrel=1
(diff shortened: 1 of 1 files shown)


community/mongodb/PKGBUILD
```diff
@@ -12,7 +12,7 @@
 
 pkgname=mongodb
 pkgver=3.2.6
-pkgrel=1
+pkgrel=2
```
kmihelich: Go back and read the rules for contributing.

TPXP:
Sorry for the previous error. It should be OK now. While I was on it, I added a note at the beginning of the file. :smile:

PR #1368, opened 2016-06-05
extra/mesa: new package mesa-wayland-egl

mickybart:
Package mesa-wayland-egl handles libwayland-egl.so library in the same way than mesa-libgl.
This permit installation of another libwayland-egl.so provider like libhybris.

For transparency, mesa-wayland-egl is a dependency of mesa-libgl.

### DIFF
--- a/extra/mesa/PKGBUILD
+++ b/extra/mesa/PKGBUILD
@@ -9,8 +9,12 @@
 #  - Build vc4 gallium driver for v6/v7
 #  - Keep prepare function for older llvm (remove when llvm is fixed)
 
+# ALARM: Michael Serpieri <mickybart@pygoscelis.org>
+#  - Move libwayland-egl library to a new package mesa-wayland-egl
+#    to support different providers for libwayland-egl (eg: libhybris) in the same way than libgl
+
 pkgbase=mesa
-pkgname=('mesa' 'mesa-libgl' 'libva-mesa-driver')
+pkgname=('mesa' 'mesa-libgl' 'mesa-wayland-egl' 'libva-mesa-driver')
 pkgver=11.2.2
 pkgrel=1.1
 arch=('i686' 'x86_64')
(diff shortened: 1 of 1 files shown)


kmihelich:
What's wrong with using ld.so.conf.d?

mickybart:
In fact I'm not using ld.so.conf.d for few reasons:

- Developers of kwin / plasma mobile them self have got some issues mainly with wayland client application : see https://blog.martin-graesslin.com/blog/2015/11/upgrading-libhybris/ (and search for libwayland-egl)

- A bug track is opened on Ubuntu (https://bugs.launchpad.net/ubuntu/+source/mesa/+bug/1507584) with a request to move libwayland-egl.so to mesa-libgl. But on my side I prefer to introduce a new package because some proprietary blob can not include this lib like nvidia-libgl.

- I can't set the path to libhybris lib directly because there is libgl lib on it too. For those who want to try freedreno or other opensource drivers, it is important to be able to switch between mesa-libgl or libhybris-libgl

- pkgconfig (wayland-egl.pc) is not the same between mesa and libhybris. (so qt5-wayland can be compiled against libhybris)

So by the fact that we need to replace libwayland-egl.so* and wayland-egl.pc it seems better in my point of view to introduce this new package. That permit to follow the same logic than with libgl and to have a flexible solution.

I have tested a lot this setup, and I'm able to use kwin and wayland client application with libhybris under my smartphone.

kmihelich:
I don't see any reason why a combination of ld.so.conf.d and compile flags can't work; this is done in many other situations like this.

Additionally, libhybris is not a supported package here, and I can only assume you want something like this for the blast of packages you just added to the AUR.  There is no history to prove that you will be reliable in continuing this, nor do I see any active use case.

I also don't have the time to maintain an arbitrary library move between packages, and I really don't have the time to deal with the inevitable fallout with a move in a package as heavily used as mesa.

mickybart:
Ok I understand your point. Thanks

PR #1374, opened 2016-07-01
[linux-odroid-c1] Install backports headers

Xenopathic:
The kernel is compiled with the backports tree enabled (previously known as wireless-compat, it's a backport of newer networking code to older kernels). However, headers were not being copied across, so compiling external modules would fail, often with 'Unknown symbol' when modprobing (abperiasamy/rtl8812AU_8821AU_linux#42). This patch copies those headers into the -headers package.

It's likely that other architectures, certainly at least the Hardkernel ones, also need this patch, but I've only submitted for Odroid C1 since that is the platform I'm running.

### DIFF
--- a/core/linux-odroid-c1/PKGBUILD
+++ b/core/linux-odroid-c1/PKGBUILD
@@ -151,6 +151,9 @@ _package-headers() {
     media net pcmcia scsi sound trace uapi video xen; do
     cp -a include/${i} "${pkgdir}/usr/src/linux-${_kernver}/include/"
   done
+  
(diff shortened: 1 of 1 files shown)


kmihelich:
Seeing that only some networking related backports are enabled in the config, clobbering unrelated headers from the main source is not appropriate.  You need to determine the exact headers that need to be replaced given what is installed into the headers package in the first place.

Xenopathic:
I'm no longer interested in pursuing this PR, sorry.

PR #1376, opened 2016-07-08
fixed passthrough in kodi when using hifiberry-digi+ card

nornic:
The HiFiBerry Digi+ is a S/PDIF/Toslink output board for the Raspberry Pi (rPi2 and rPi3).

For years (I found forum topics going back to 2014) kodi is not able to use it as pass-through device for audio output. What that means is, you are not able to get DTS or DolbyD signal in your amp.

It looks OpenELEC and Osmc already applied the patch, or rather workaround for that issue.
More info here:
https://support.hifiberry.com/hc/en-us/community/posts/201845791-Kodi-no-passthrough
and here
http://forum.kodi.tv/showthread.php?tid=261778

Kodi won't apply the patch as in their eyes the problem should be fixed in kernel or by Hifiberry.
I really doubt it will be fixed upstream in the near future (judging on the sheer amount of topics/issues which are open all over the web).

It's worth applying that patch to save time and headache to arm arch users.

It works perfectly, without side effects, in my home setup for quite some time now.

### DIFF
--- a/alarm/kodi-rbp/PKGBUILD
+++ b/alarm/kodi-rbp/PKGBUILD
@@ -12,7 +12,7 @@ pkgbase=kodi-rbp
 pkgname=('kodi-rbp' 'kodi-rbp-eventclients')
 pkgver=16.1
 _codename=Jarvis
-pkgrel=3
+pkgrel=4
 pkgdesc="A software media player and entertainment hub for digital media for the Raspberry Pi"
 arch=('armv6h' 'armv7h')
 url="http://kodi.tv"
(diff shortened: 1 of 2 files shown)


moonman:
Please read https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md for future PRs
specifically the title should state the package being updated.
Changes in this PR committed.

nornic:
will do.

Thanks for merging

hansheijmans:
Compiled and installed using makepkg -s. Took me hours to compile on RPi2 but still no HifiBerry Digi+ I can select!!!

What am I doing wrong??

Please help!

nornic:
@hansheijmans: the patch only adds digi+ spdif/toslink support in kodi. Card should work in previous kodi builds, in analog mode only.

If you can't see hifiberry card at all, then you haven't set it up correctly. 
That's the guide which helped me: https://www.hifiberry.com/guides/configuring-linux-3-18-x/

AFAIR set-up is as simple as adding to /boot/config.txt:
`dtoverlay=hifiberry-digi`
creating /etc/asound.conf:
```
pcm.!default  {
 type hw card 0
}
ctl.!default {
 type hw card 0
}
```
and rebooting pi. That's it, you should be able to see it as a separate audio device in kodi.

`aplay -L` should list you: `snd_rpi_hifiberry_digi`
if you can see it but no sound comes out, then remember about setting outputs and un-muting device in `amixer`.

Hope that helps

hansheijmans:
Thanks for your reply!

I forgot to mention that I already included dtoverlay=hifiberry-digi in the config.txt as I am using this card for other purposes as well successfully (so aplay -l also lists this card) e.g. with MPD, spopd, shairport-sync.

When using Open/LibreElec/OSMC etc. I can also see and use the card.

Point is I want to install kodi on a standard ArchLinux Arm environment (for RPi2) which also hosts other applications. With kodi-rbp from the repos I can't use the Hifiberry card which is not supported by the un-patched kodi from the repos (not by this version and also not by previous versions).

My guess is that something went wrong with the makepkg procedure. I cloned the complete PKGBUILDs tree to my local SD-card on the RPi2 and started makepkg -g and makepkg -s from within the kodi-rbp directory which also contains the patches. With pacman -U I installed both created tarballs and started kodi with "systemctl start kodi". I even tried "AE_SINK=ALSA kodi" but that didn't help either, still no Hifiberry Digi+. I must say that during the build I came across zillions of warnings about conflicting cpu architecture parameters (-march=cortex-a7 vs. armv7-a as far as I can remember).

So my question remains the same...and maybe you know how to speed up building with makepkg as my first attempt took about 4 hours to complete.

moonman:
Please move this conversation to https://archlinuxarm.org/forum/
as this is not related to the PR.

PR #1379, opened 2016-07-20
Add getty.target to kodi systemd service

NicoHood:
When running the kodi systemd service at boot and you play a movie with blackborders (or wait for the screenscaver dimm) you can see the console login still in the background. With a manual start of the service this does not happen. You need to wait for the getty to finish.

The service was tested with and without the setting multiple times to veryfy that it works.
This patch should possibly get applied to kodi-rbp-git too (not tried).

### DIFF
--- a/alarm/kodi-rbp/kodi.service
+++ b/alarm/kodi-rbp/kodi.service
@@ -1,6 +1,6 @@
 [Unit]
 Description = Starts an instance of Kodi
-After = remote-fs.target
+After = remote-fs.target getty.target
 
 [Service]
 User = kodi


NicoHood:
You were 2 minutes too fast, git did not catch my ammended commit. I updated the packagerel of course. I do not know how github handles closed PRs with updates.

moonman:
Don't know if you updated this, but the title should state the package being updated. You have "kodi" there but there are several kodi packages. Take a look at the repository commit logs (/alarm/kodi-rbp: blah blah)

NicoHood:
Oh you are totally right. I missed that. I've added this to my commit message now via a force push. This will not get displayed on github until the pr is reopened i think.
You can see the changes here: https://github.com/archlinuxarm/PKGBUILDs/compare/master...NicoHood:patch-2?expand=1

NicoHood:
Is there anything I can do? My current branch should be okay now:
https://github.com/archlinuxarm/PKGBUILDs/compare/master...NicoHood:patch-2?expand=1

I'd be really happy if this bug can get fixed. This should be also tested for the alarm/kodi-rbp-git package as it might also have this bug. I've also fixed my other PR about wiringpi. So what else may I do?

Thanks

NicoHood:
`gdm` can also be added as optional dependency. This will provide a more detailed debug log with stack trace.

PR #1382, opened 2016-08-16
core/linux-odroid-xu3: Enable kernel modules for CAN bus and devices

normaldotcom:
This will add support for CAN bus connectivity and CAN bus devices, which are very common in situations where embedded hardware is used.


moonman:
This kernel doesn't compile with gcc6 at the moment.

normaldotcom:
I have a trivial patch for /arch/arm/mach-exynos/exynos-ss.c which fixes compilation under gcc6. Should I add this patch in a new PR? Patch included below.

`
0522787bdde1248bd/arch/arm/mach-exynos/exynos-ss.c
241,247d240
< enum ess_cause_emerg_events {
<       CAUSE_INVALID_DUMP = 0x00000000,
<       CAUSE_KERNEL_PANIC = 0x00000001,
<       CAUSE_FORCE_DUMP   = 0x0000000D,
<       CAUSE_FORCE_REBOOT = 0x000000FF,
< };
<
502a496,501
> enum ess_cause_emerg_events {
>       CAUSE_INVALID_DUMP = 0x00000000,
>       CAUSE_KERNEL_PANIC = 0x00000001,
>       CAUSE_FORCE_DUMP   = 0x0000000D,
>       CAUSE_FORCE_REBOOT = 0x000000FF,
> };
`

WarheadsSE:
@normaldotcom any patch such as that should be separated from a change to the config, unless that change is to enable the compatibility with GCC v#, which is not going to be the case.

PR #1383, opened 2016-08-17
core/linux-odroid-xu3: Add patch to exynos-ss for build under gcc6

normaldotcom:
This kernel currently doesn't build under gcc6 due to the position of an enum declaration in `exynos-ss.c`. This patch resolves the issue, and the kernel builds successfully under gcc 6.1.1.


WarheadsSE:
@mdrjr thoughts?

moonman:
@normaldotcom thanks for figuring this out

PR #1384, opened 2016-09-01
extra/mesa: Build vc4 gallium driver on aarch64

sirspudd:
This change extends sha # de39dc2, adding aarch64 to the array of
platforms which build the Raspberry Pi's vc4 driver. I have tested the
driver on the Arch aarch64 rootfs and can confirm that it is functional
and usable in its current state.

This patch should be merged so that when a Raspberry Pi 3 is booted with a functional kernel and the Arch aarch64 rootfs, mesa should be immediately usable for EGL/OpenGL ES 2 applications. It present one has to rebuild mesa in order to successfully get mesa based hardware acceleration on the Raspberry Pi 3.

### DIFF
--- a/extra/mesa/PKGBUILD
+++ b/extra/mesa/PKGBUILD
@@ -40,7 +40,7 @@ prepare() {
 build() {
   cd ${srcdir}/?esa-*
 
-  [[ $CARCH == "armv7h" || $CARCH == "armv6h" ]] && VC4=',vc4'
+  [[ $CARCH == "armv7h" || $CARCH == "armv6h" || $CARCH == "aarch64" ]] && VC4=',vc4'
 
   ./configure --prefix=/usr \
     --sysconfdir=/etc \


sirspudd:
I am not quite sure of the history regarding the original exclusion of vc4 on certain architectures, but it functionally spans 3 architectures (alot more than many of the included drivers, like nouveau) and works perfectly on aarch64

WarheadsSE:

sirspudd:
@WarheadsSE My apologies for not initially following those contribution guidelines closely enough, hopefully this commit message is closer to what is required. I also adjusted the change to be more inline with what is already there, although I am not certain why vc4 is being treated any differently from the other gallium drivers which are explicitly specified in the configure line. My preference would be to simply include vc4 in the existing array, but I assume there is some legacy reason for excluding the building of this driver on certain platforms, which I defer to.

WarheadsSE:
Please update the PR title and comment to reflect the requirements of the contribution guidelines.

yangxuan8282:
@sirspudd can we use OpenGL on RPi3 with archlinux arm aarch64 ?

sirspudd:
@yangxuan8282 Yes, I have tested it on aarch64 with both Fedora (where it is built with VC4 out of the box) and on Arch after explicitly rebuilding mesa with VC4 support.

yangxuan8282:
@sirspudd How about performance? Is it good as on raspbian?

sirspudd:
For simple OpenGL accelerated QML examples there is no appreciable difference between the proprietary GL drivers and vc4, that I can see. It runs really well.

yangxuan8282:
@sirspudd Would you mind share the instruction about how to get rpi3 running with aarch64 archlinux, or some links?

sirspudd:
@yangxuan8282 Merrily:

http://chaos-reins.com/2016-09-01-qt-pi3-arch-aarch64/

is a blog I wrote about the process. Basically Fedora alomst works out the box, requiring just a different device overlay, and then Arch worked almost out the box (with the Fedora kernel), except for the need to recompile mesa with vc4 support.

yangxuan8282:
@sirspudd 
Seems archlinuxarm finally release aarch64 system for RPi3. I have installed it yesterday, not sure if miss anything, enter xfce desktop spend 1 minutes, after waiting 2 minutes I can open lxterminal (which usually take less than 20 seconds on ARMv7 system). The keyboard act is wired sometime, the system think I'm keep pressing on some key, so it output like "aaaaaaaaaaaaaa", after reboot, got "Could not resolve host error" issues, add google nameserver in `resolv.conf` didn't solve the problem, so I guess some other files is corrupted.

PR #1391, opened 2016-09-22
alarm/uboot-odroid: Use generic distro configuration

ztombol:
_This is a **WORK IN PROGRESS**. I wanted to gauge interest before I gather volunteers on the forums to test on boards that I don't own and fully test everything._

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

I'm aware that U-Boot's *generic distro config* has been [discussed before][alarm-prev-pr] and that at that time it wasn't accepted. I'm wondering if the stance has changed. Especially in light that in the meanwhile U-Boot has gained a flexible common implementation that supports falling back to a `boot.scr` for backward compatibility.

Merit:
- **Less code to maintain.** U-Boot implements the boot environment which is shared between all boards that enable this configuration method. Boards only have to define the memory addresses where various images should be loaded and the devices which should be searched for boot files. Modifying device order, boot prefixes, and customising network setup is supported too. This results in a smaller Arch Linux ARM specific patch.

- **Flexible runtime.** All partitions marked as bootable are searched for boot files. Defaults to the first partition if no such partition is found. Boot files may reside in `/` or `/boot/`, which allows using separate `boor` and `root` partitions out-of-the-box. Falls back to `boot.scr` if `extlinux.conf` is not found or cannot boot.
  
- **Simpler configuration.** _Some of this may be subjective._ The path of the kernel, initial ramdisk and device tree can be easily specified without having to look at the compiled-in environment. Kernel arguments are specified on a single line, rather than being concatenated from a number of environment variables, while the compiled-in environment adds a few more. Multiple configurations can be specified in a single file, and switching between them is as simple as editing one line.

In my opinion the main benefit is the user friendly, as in more familiar and transparent, configuration and the reduced maintenance cost. The expanded feature set is nice too, but I know it isn't important for Arch Linux ARM.

See [`doc/README.distro`][uboot-readme-distro] in U-Boot's source for more details. [My U-Boot branch][ztombol-uboot-odroid] implementing this for the boards supported by the `uboot-odroid` package.

Alternatives:
- If `extlinux.conf` is not something you like, it's still possible to reduce maintenance cost by utilising the boot file searching logic only and falling back to an Arch Linux ARM specific boot script. This saves only a couple of lines of code in this package. However, using the *generic distro config* provides the boot script a standard environment, which enables writing a boot script that is generic enough to be shared among other boards and thus reducing maintenance cost across multiple platforms. The second commit introduces such a script (still under testing). It supports Flattened Image Trees (`Image.itb`), compressed kernel images (`zImage`), and legacy `uImage`, as well as flattened device trees and initial ramdisks.

Implementation notes:
- As of `2016.07`, USB and Etherent (possibly because it's connected via USB) are broken in upstream U-Boot. On my U2, I couldn't bring up network as described in [`doc/README.odroid`][uboot-readme-odroid] and couldn't boot stock Arch Linux ARM from a USB drive either. Only eMMC and SD card boot is enabled at the moment.
- [`include/config_distro_defaults.h`][uboot-distro-defaults-h] already contains `CONFIG_SUPPORT_RAW_INITRD`, so it's not in the boards header.
- U-Boot [loads `extlinux.conf` at `scriptaddr`][uboot-extlinux-load] instead of [`pxefile_addr_r` as the documentation says][uboot-pxefile_addr_r]. This is a bug, but it doesn't matter, especially that in case of these boards those addresses are the same.
- `${fdtfile}` is automatically set for all boards (`exynos4412-odroid{x,x2,u2,u3}`). There's no need to set it to the empty string.

References:
- 2014-07-18 - [Previous discussion in #907][alarm-prev-pr]
- 2014-07-31 - [U-Boot adds generic $bootcmd][uboot-gen-bootcmd]
- 2015-01-22 - [U-Boot documents generic distro config][uboot-add-readme-distro]

[alarm-prev-pr]: https://github.com/archlinuxarm/PKGBUILDs/pull/907
[uboot-readme-distro]: https://github.com/u-boot/u-boot/blob/v2016.07/doc/README.distro
[uboot-readme-odroid]: https://github.com/u-boot/u-boot/blob/v2016.07/doc/README.odroid
[ztombol-uboot-odroid]: https://github.com/ztombol/u-boot/tree/alarm-uboot-odroid-extlinux
[uboot-extlinux-load]: https://github.com/u-boot/u-boot/blob/v2016.07/include/config_distro_bootcmd.h#L336
[uboot-pxefile_addr_r]: https://github.com/u-boot/u-boot/blob/v2016.07/doc/README.distro#L252
[uboot-gen-bootcmd]: https://github.com/u-boot/u-boot/commit/2a43201a13bc690b5f15a1a998514d932db2f1b0
[uboot-add-readme-distro]: https://github.com/u-boot/u-boot/commit/ffb4f6f95a0b63a0e8eab81e006a26c9cd99ac5d
[uboot-distro-defaults-h]: https://github.com/u-boot/u-boot/blob/v2016.07/include/config_distro_defaults.h#L35

### DIFF
--- a/alarm/uboot-odroid/0001-arch-linux-arm-modifications.patch
+++ b/alarm/uboot-odroid/0001-arch-linux-arm-modifications.patch
@@ -1,14 +1,24 @@
-From def7823985d4a7b7670c1ed6eea8a28482ae4642 Mon Sep 17 00:00:00 2001
-From: Kevin Mihelich <kevin@archlinuxarm.org>
-Date: Fri, 22 Aug 2014 17:51:41 -0600
-Subject: [PATCH 1/2] arch linux arm modifications
+From 667549a1624dc76e631a0db1b3abc6fee115e80e Mon Sep 17 00:00:00 2001
+From: Zoltan Tombol <zoltan.tombol@gmail.com>
+Date: Mon, 19 Sep 2016 19:42:57 +0200
+Subject: [PATCH] odroid: Add generic distro configuration
 
 ---
- include/configs/odroid.h | 173 ++++++++++++++++++-----------------------------
- 1 file changed, 64 insertions(+), 109 deletions(-)
+ configs/odroid_defconfig |   1 +
+ include/configs/odroid.h | 142 +++++++++++------------------------------------
+ 2 files changed, 34 insertions(+), 109 deletions(-)
 
+diff --git a/configs/odroid_defconfig b/configs/odroid_defconfig
+index 76ab144..f994f2a 100644
+--- a/configs/odroid_defconfig
++++ b/configs/odroid_defconfig
+@@ -47,3 +47,4 @@ CONFIG_G_DNL_MANUFACTURER="Samsung"
+ CONFIG_G_DNL_VENDOR_NUM=0x04e8
+ CONFIG_G_DNL_PRODUCT_NUM=0x6601
+ CONFIG_ERRNO_STR=y
++CONFIG_DISTRO_DEFAULTS=y
 diff --git a/include/configs/odroid.h b/include/configs/odroid.h
-index 12b7dc5..306ff07 100644
+index 12b7dc5..35c4ab6 100644
 --- a/include/configs/odroid.h
 +++ b/include/configs/odroid.h
 @@ -14,6 +14,7 @@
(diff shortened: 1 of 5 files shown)


kmihelich:
No.

If you want to switch the odroid config to the distro default setup, send a patch upstream to change it.  Then in the future this package will switch to the standard boot.scr setup used for other boards, which is far more flexible, simple, and easier to maintain than the extlinux cruft.

Given your lack of previous contributions to this project, history tells me you intend to dump this on me to maintain going forward while you disappear.  That will not happen.  This is not less to maintain from both packaging and support perspectives, and only serves to unnecessarily complicate and obfuscate the boot process.

ztombol:
@kmihelich 

No. I didn't intend to dump it on you and disappear. Everyone has to start somewhere.

I agree that upstream would be the best place for this, but that would require removing/replacing obsolete implementation for these boards, e.g. see the I2C warning during compilation and the memory test in [`doc/feature-removal-schedule.txt`][doc-feature-rem]. I don't have the necessary knowledge to do this.

The second best place would be a distro, which could benefit from less code.

I'm willing to work with you if you want `boot.scr` to be simpler or replace it with one you already use with other boards. Most of my pull request is documentation which can be tailored to your liking.

> standard boot.scr setup used for other boards, which is far more flexible, simple, and easier to maintain than the extlinux cruft.

The extlinux implementation is done by upstream. The only thing you need to maintain is a config file.

The first commit adding extlinux support changes less of the boards header than the current patch. The rest of the commit is a well documented configuration file.

Knowing that you most likely won't like the `extlinux.conf` way, the second commit adds a generic `boot.scr` and `uEnv.txt` for demonstration. They can be replaced with whatever you already use for other boards.

This would allow you to move most of the changes out from the boards' header to a `boot.scr` that you then can share between other boards. I don't see how this is more than what you already have.

[doc-feature-rem]: https://github.com/u-boot/u-boot/blob/v2016.07/doc/feature-removal-schedule.txt

PR #1395, opened 2016-10-08
Enable 32bit support in linux-aarch64

yuyichao:
I've asked about this on the IRC and on the [forum](https://archlinuxarm.org/forum/viewtopic.php?f=65&t=10829) but got no reply so I guess I'll just submit a PR and try to get feedback this way.

This PR doesn't fix all the problems I have with the kernel but it at least makes it possible to run 32bits applications with the stock kernel. The only flags I manually changed are the ones mentioned in https://community.arm.com/message/38647#38647. All other ones are dependencies.


yuyichao:
For the record 64k page was enabled in https://github.com/archlinuxarm/PKGBUILDs/commit/6702e6123fc1d982b52a7a48e794bb3a5b6ed5a6 which removes `COMPAT`, which wasn't enabled, and causes issues with many armv7h binaries in the repo. @kmihelich is it intentional?

kmihelich:
Yes, this was intentional.  This kernel is for the aarch64 repo, which you should be using, not armv7h.

yuyichao:
So what kernel should I use if I want to use 32bit container on a 64bit host?

yuyichao:
So what about another kernel that has the support? Or do you mean that this is a feature that does make sense to support in the official repo?

kmihelich:
No, I won't be maintaining yet another kernel package for this.

yuyichao:
@kmihelich FYI, I think none of the patches are actually applied.

yuyichao:
The last bit of 32bit support in the 64bit kernel is finally accepted at https://github.com/archlinuxarm/PKGBUILDs/pull/1929 .

PR #1400, opened 2016-10-24
alarm/kodi-odroid new package (17.0.beta4)

frol:
I have been heavily testing it on my Odroid U3 for over 2 weeks and it works great for me.

Since all dependencies are Odroid X/X2/U2/U3 specific, I suppose, it should work on those devices as well.

The PKGBUILD is based on kodi-c1 PKGBUILD.

NOTE: MFC firmware couldn't have been found on my Odroid U3 because the `s5p-mfc*.fw` files are installed into `/lib/firmware/`, but once I symlinked these `*.fw` files to `/lib/firmware/s5p-mfc/` folder, it worked like a charm (see details here: https://archlinuxarm.org/forum/viewtopic.php?f=60&t=10913). This issue is not related to the packaging, but it is related to the user experience.

### DIFF
--- a/alarm/kodi-odroid/PKGBUILD
+++ b/alarm/kodi-odroid/PKGBUILD
@@ -1,0 +1,136 @@
+# Maintainer: Kevin Mihelich <kevin@archlinuxarm.org>
+# Maintainer: Oleg Rakhmanov <oleg@archlinuxarm.org>
+# Contributor: Sergej Pupykin <pupykin.s+arch@gmail.com>
+# Contributor: BlackIkeEagle < ike DOT devolder AT gmail DOT com >
+# Contributor: Brad Fanella <bradfanella@archlinux.us>
+# Contributor: [vEX] <niechift.dot.vex.at.gmail.dot.com>
+# Contributor: Zeqadious <zeqadious.at.gmail.dot.com>
+# Contributor: BlackIkeEagle < ike DOT devolder AT gmail DOT com >
+# Contributor: Bartłomiej Piotrowski <bpiotrowski@archlinux.org>
+# Contributor: Maxime Gauduin <alucryd@gmail.com>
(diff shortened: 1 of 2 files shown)


frol:
NOTE: Another issue I faced was that `/dev/mali` (video card device) and `/dev/CEC` ([Consumer Electronics Control](https://en.wikipedia.org/wiki/Consumer_Electronics_Control) device) had `root:root rw-rw----` permissions, so users should create and add themselves into `video` group, and add udev rules to make the permissions sane:

```
# /etc/udev/rules.d/50-mali.rules
KERNEL=="mali", MODE="0660", GROUP="video"
```

```
# /etc/udev/rules.d/51-CEC.rules
KERNEL=="CEC", MODE="0660", GROUP="video"
```

kmihelich:
I'll make some package updates to deal with the s5p-mfc firmware and Mali rules, but these devices are going to be switched to the mainline kernel in the near future, which doesn't have Mali support.  The 3.8 kernel will be left for some period of time for those that really want to use it, but it's very much EOL and will receive little or no continued support.

Supporting this package will be entirely on you.  If you abandon it here, it will be removed.  If you agree to this I'll consider this PR.  You might also consider just putting this in the AUR, as I know of no expressed demand for Kodi on these boards.

frol:
> these devices are going to be switched to the mainline kernel in the near future, which doesn't have Mali support.

I would like to share that @ztombol [prepared PKGBUILDs](https://github.com/ztombol/odroid-tools) for @tobiasjakobi's [mainline kernel fork](https://github.com/tobiasjakobi/linux-odroid-public) (currently, it is quite up-to-date 4.8 kernel with Mali support).

I have seen some demand on the forum, and I agree to maintain this package while I have my Odroid U3.

Thank you.

PR #1412, opened 2016-11-12
alarm/kodi-rbp to 16.1-5

NicoHood:
* Fixes transparent black boarders when used as systemd service
* Added gdb as optional dependency for stack traces
* Start kodi after lircd service to ensure lircd runs first
* Fixed a polkit permission warning when installing the package

-> Those changes should also be adopted to the -git package possibly.


NicoHood:
Thanks for merging. I am not sure why the install note is there, but it seems to be not relevant unless this is a special case:
`echo 'SUBSYSTEM=="vchiq",GROUP="video",MODE="0660"' > /etc/udev/rules.d/10-vchiq-permissions.rules`

PR #1415, opened 2016-11-15
community/python-kivy: search /usr/lib for libMali

jakogut:
When detecting platform 'mali', the current Kivy setup code uses a
predefined path to detect the presence of the userspace mali GLES
library, which works on Debian based distros.

Other distros may install the library in a different location. Arch,
for instance, installs it to /usr/lib/mali-egl/libMali.so.

This patch searches /usr/lib/ recursively for the library.


kmihelich:
Mali libraries are not present in the build environment.

jakogut:
@kmihelich Hmm, that's tricky. What's the solution? A separate package that makedepends a specific Mali userspace package?

kmihelich:
Get the developer to stop doing compile time checks and instead do runtime checks. I don't want separate packages here for every variant.

jakogut:
That sounds like the correct technical approach.

PR #1417, opened 2016-11-22
alarm/kodi-rpb-git: switch to software decode for GMC files

graysky2:
Changes introduced:

* Include patches from popcornmix's branch of kodi to solve issue of video tearing on playback of some video files.

Why needed:
* VIdeo files encoded with GMC (global motion compensation) cause the problem on RPi3.  The patches switch to software decoding mode when GMC is detected to avoid the tearing without user intervention and are including in LibreELEC and OSMC.

> Due to the extra decoding CPU cost of global motion compensation, most hardware players do not support global motion compensation.

Verified to solve the tearing issue on my box; non-GMC files seem to play just fine for the 5 that I tested.

[Link](http://forum.kodi.tv/showthread.php?tid=297524) to kodi forum thread for additional background info.

### DIFF
--- a/alarm/kodi-rbp-git/PKGBUILD
+++ b/alarm/kodi-rbp-git/PKGBUILD
@@ -12,7 +12,7 @@ pkgname=('kodi-rbp-git' 'kodi-rbp-git-eventclients')
 pkgver=17.0b5.20161029
 
 _tag=17.0b5-Krypton
-pkgrel=4
+pkgrel=5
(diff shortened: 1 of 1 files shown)


moonman:
Will RPI 1 handle software decoding or will putting up with tearing be better than choppy playback? Should it be excluded (i.e. only patch builds for armv7 )?

graysky2:
@moonman - Good question.  I no longer have a RPi only RPi2 and RPi3 so I cannot test it.  If you have a RPi I can provide you with a test video that does tear due to GMC.  Please let me know.

PR #1419, opened 2016-12-06
odroid-c2: added xf86-video-fbturbo-odroid-c2-git package

kroylar:
Currently, this fbturbo driver is faster than the mali driver on odroid-c2. I added a new package entirely based on the [current xf86-video-fbturbo-git PKGBUILD](https://github.com/archlinuxarm/PKGBUILDs/blob/master/alarm/xf86-video-fbturbo-git/PKGBUILD).


kmihelich:
This doesn't need to be built for all architectures.

And this package will most definitely conflict with an xf86-video-fbturbo-git installation, which you are not accounting for.

And a typo in your pkgdesc.

kroylar:
Thank you for your guidance Kevin. I think I've addressed all three of your comments.

PR #1424, opened 2016-12-18
Bump libbcm2835 version and enable a64 builds

ilpianista:


kmihelich:
You need up update the buildarch variable too, and squash your commits.

ilpianista:
up!

PR #1429, opened 2016-12-28
Add xf86-video-armsoc-sunxi

ilpianista:
[longsleep](https://github.com/longsleep) [patched armsoc](https://github.com/longsleep/xf86-video-armsoc) for sunxi. I would like to get it on Arch Linux ARM if possible.


kmihelich:
I don't see how this is applicable to any currently supported device.

ilpianista:
Ok, point taken.

This makes no sense until ArchLinux ARM supports Pine64 officially.

kmihelich:
The mainline kernel doesn't support Mali, so it would make no sense then either.

PR #1430, opened 2016-12-29
Proper support of HDMI CEC on ODroid-C2

mdilai:
Hello,

Currently ALARM kernel linux-odroid-c2 are compiled with old HDMI CEC driver
CONFIG_AML_AO_CEC
instead of the new one 
CONFIG_AML_HDMI_TX_NEW_CEC_DRIVER=y
This has been fixed in config.

Also libcec must be compiled with it's support. Libcec 4.0 already have AOCEC adapter support but it's expect device name /dev/aocec and virtual COM name "AOCEC". But latest Hardkernel sources using /dev/AmlogicCEC and "Amlogic" respectively. I've merged LibreElec patch for support Amlogic so both variants are supported at the same build.

This critically needed for supporting CEC on ODroid-C2.


kmihelich:
Aside from your blatant disregard of the rules for contributions, the exact same patch for libcec has been submitted upstream.  I'll wait for that to pass review and get included.

PR #1435, opened 2017-01-09
alarm/aml-libs-c2 to r9.296f39b

phedoreanu:
* `source` updated to use `${_commit}.tar.gz`
 * `sha256sums` used in favour of `md5sums`


kmihelich:
There is no functional change, and you just made this harder to maintain.  No.

PR #1436, opened 2017-01-09
extra/pyqt5: Add back qt5-webengine makedepend

ryan-farley:
With a6a83b917e609abaa29d6f46e4762792097141e7 QtWebEngine is now
functional.


kmihelich:
And this will render pyqt5 unbuildable for v5 and v6.  Not happening at this point.

PR #1440, opened 2017-01-15
alarm/kodi-rbp-git: update to 17.0rc3.20170114-1

graysky2:
Update to use upstream supported cmake.  Tested on RPi2 and works!  Lirc works too.


moonman:
I don't think removing the C[xx]flags is a good idea, especially for armv7 (rpi2) build. Default flags are for generic armv7 builds. This package is for RPI 2 specifically and the flags are there to further optimize the build for this device.

graysky2:
Added back.  Builds and runs fine.

moonman:
Ah, I see. Perfect.

PR #1444, opened 2017-01-24
alarm/uboot-sunxi: Add BananaPi support

CReimer:
I tried to get this in here before with a pull request over 2 years ago (#892)
A lot has changed since then.

@kmihelich had a very strong opinion about these boards back then. And one of the key arguments to not implement my pull request was that there was no upstream uboot support available at that time.

I actually sold my Bananapi back then, because it was too much trouble to keep it running. A few weeks ago I bought another one which serves with Arch Linux ARM as my home router. There is absolutely no problem with this hardware. It works just as good as the Cubieboard 2 right next to me (except it has a gigabit ethernet controller)

Another rather subjective argument was that it is "chinese copycat garbage". It's a lot easier to buy a BananaPi in Germany than it is to buy a Cubieboard 2. And the European Union and Germany in particular have very strict laws to prevent "chinese copycat garbage". And still these boards are sold openly by all big electronic stores.

There seemed to be two other attempts by @synthead and one by @seidler2547, which were gunned down by @kmihelich.

In one of these pull requests another argument came up. Lemaker supposedly spams your site and social media. Well, I can't blame them. You refuse to add a hilariously simple patch to support their hardware.

And to close this up, here's another interesting fact. Allnet mentions Lemaker on their website. They don't just mention them. It's more like praising them for their good products.

I'd like to keep this request's discussion as rational as possible. So could we have a discussion here where everyone states arguments based on ones opinion and lets someone else counter these arguments.

Thanks in advance

Christopher

### DIFF
--- a/alarm/uboot-sunxi/PKGBUILD
+++ b/alarm/uboot-sunxi/PKGBUILD
@@ -11,6 +11,7 @@ pkgname=('uboot-a10-olinuxino-lime'
          'uboot-a20-olinuxino-lime'
          'uboot-a20-olinuxino-lime2'
          'uboot-a20-olinuxino-micro'
+         'uboot-bananapi'
          'uboot-cubieboard'
          'uboot-cubieboard2'
          'uboot-cubietruck'
@@ -38,6 +39,7 @@ boards=('A10-OLinuXino-Lime'
         'A20-OLinuXino-Lime'
         'A20-OLinuXino-Lime2'
         'A20-OLinuXino_MICRO'
+        'Bananapi'
         'Cubieboard'
         'Cubieboard2'
         'Cubietruck'
@@ -160,6 +162,20 @@ package_uboot-a20-olinuxino-micro() {
 
   install -d "${pkgdir}"/boot
   install -Dm644 bin_A20-OLinuXino_MICRO/u-boot-sunxi-with-spl.bin "${pkgdir}"/boot
+
+  install -Dm644 boot.txt "${pkgdir}"/boot/boot.txt
+  install -Dm644 boot.scr "${pkgdir}"/boot/boot.scr
+  install -Dm755 mkscr "${pkgdir}"/boot/mkscr
+}
+
+package_uboot-bananapi() {
+  pkgdesc="U-Boot for BananaPi"
+  install=${pkgbase}.install
+  provides=('uboot-sunxi')
+  conflicts=('uboot-sunxi')
+
+  install -d "${pkgdir}"/boot
+  install -Dm644 bin_Bananapi/u-boot-sunxi-with-spl.bin "${pkgdir}"/boot/u-boot-sunxi-with-spl.bin
 
   install -Dm644 boot.txt "${pkgdir}"/boot/boot.txt
   install -Dm644 boot.scr "${pkgdir}"/boot/boot.scr


seidler2547:
👍 🍌

kmihelich:
Nothing's changed; still not a candidate for official support.

CReimer:
And why is it still no candidate? I tried to counter all your previous arguments and in my opinion none of them are valid anymore.

And yes, a lot has changed.
What am I missing? It just doesn't make any sense to me. Please, explain it.

moonman:
Regarding chinese copycat... it is.
- The name
- Go to Banana Pi official website and then visit Intel website. Compare the logos. These are just two things I noticed.

CReimer:
The name. Yes it's similar to Raspberry Pi. The Raspberry Pi Foundation doesn't seem to be concerned about this, why should Arch Linux ARM be?

The logo: In my opinion this is somewhat far-fetched. But again, I don't see a reason why the Arch Linux ARM project should be concerned about this. That's Intel's thing. If they think it's a problem then it's their problem and not the problem for anyone here.

When we're talking about potential Rasperry Pi clones, what about the Asus Tinkerboard? Asus has a good reputation and it is also somewhat Chinese (actually Taiwanese, but for the sake of it).

If you enforce these rules onto the Banana Pi, then there should never be any support for the Asus Tinkerboard in Arch Linux ARM

kmihelich:
There are no plans to support the Tinkerboard either.  There are far more interesting problems and systems to spend my time and money on than this month's latest uninspired Pi clone.  When one of those companies is capable of doing something original and can spend 30 seconds to come up with a name for the board that doesn't include "Pi," maybe something will happen.

synthead:
> When one of those companies is capable of doing something original and can spend 30 seconds to come up with a name for the board that doesn't include "Pi," maybe something will happen.

This is so unreasonably petty.  Instead of dictating your unpopular opinion in such a way where the entire community suffers, why not allow open software to function openly?  If you exercised nuance more, you would understand that you have your opinions for your very good reasons, but others also have the ability to believe something completely different.

Users should decide whether or not to support Banana Pi on their own without someone telling them what to think.  Unfortunately, for some people, the importance of this is not fully understood until it is taken away from them, [which is what you are actively doing to your users](https://www.reddit.com/r/linux/comments/3qt7ky/thought_id_share_my_unfortunate_interaction_while/).

CReimer:
> There are far more interesting problems and systems to spend my time and money on than this month's latest uninspired Pi clone.
Who said you should spend time and money on it. You've got three Banana Pi owners here, wo tell you that my Pull Request works. This whole topic could have been resolved with just one click in Github's webinterface.

Noone expects you to buy all hardware and test them.

I'm starting to think there's an ulterior motive here, which led to such strong opinions about one particular vendor.

kmihelich:
I am speaking specifically to what is or will be officially supported.  The community is free to do whatever it wants outside of official support, and many people do this already for a number of boards.  Merely including the package here would imply official support to the vast majority of users despite any 72pt blinking red warnings we could put up state otherwise.  This then leads to said users expecting us to fix their problems.

This position is no different than it's been throughout the life of this project.  Official support means an active, contributing developer has access to the hardware and can actively develop packages and address issues for that target.  The hardware that we choose to spend our time and money on is our own choice, not a decision forced upon us.

CReimer:
We're talking about a simple configuration setting in uboot. It either works or it doesn't. Either way, it's supported by upstream uboot and therefore any possible bug is automatically an upstream bug. You can just tell people to report it upstream and you're done.

That's exactly what the "Reporting bug guidelines" of Arch Linux say. 
> If Arch is not responsible for a bug, the problem will not be solved by reporting the bug to Arch developers.

https://wiki.archlinux.org/index.php/Reporting_bug_guidelines

pepedog:
This is a big row, and I am vodka fuelled, I read everything but even in this state I notice you haven't incremented pkgrls or taken responsibility as maintainer.
@kmihelich doesn't deserve this, he does most of the graft whilst also doing a real job too.

CReimer:
Yes, I missed that. It's fixed now.

PR #1457, opened 2017-02-23
aur/dateutils new package from AUR

kikadf:
I tested on armv7h, built succesfull and works fine. If you merge, I can maintaining the package. Of course, if you would like that ;)


kmihelich:
Not interested.

PR #1466, opened 2017-03-20
Add package for FriendlyARM NanoPI Neo to uboot-sunxi. 

rwtmck:
- Add package for FriendlyARM NanPI Neo to uboot-sunxi.
- Update to uboot-2017.03 in the process.
- The package will include the *.dtb build together with the *.bin, overwriting whatever is already put into /boot/dtbs/ by the image, because that actually *works*. Why are the *.dtb not shipped with the uboot packages?
- Tested on two different NanoPI Neo 512MB.


kmihelich:
Not a supported platform.

rwtmck:
Why isn't it supported?

PR #1472, opened 2017-04-06
Include patch from upstream to fix font rendering on some ARM boards

bhush9:
See: https://codereview.qt-project.org/#/c/185259/

This patch likely won't be included in the archlinux main repo currently
so submitting this here

### DIFF
--- a/extra/qt5-base/PKGBUILD
+++ b/extra/qt5-base/PKGBUILD
@@ -10,7 +10,7 @@
 pkgname=qt5-base
 _qtver=5.8.0
 pkgver=${_qtver/-/}
-pkgrel=7
+pkgrel=7.1
 arch=('i686' 'x86_64')
 url='http://qt-project.org/'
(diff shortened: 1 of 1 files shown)


kmihelich:
What have you tested this on where you see an improvement?

bhush9:
I've tested this patch on Nexus 5X earlier along with libhybris, and Plasma mobile.. Without this patch font rendering in qtquick Text items was broken, while this patch fixes it

kmihelich:
So nothing supported by this project.  It appears to have been merged, so it'll be there in the next version.

bhush9:
It unfortunately won't be available soon, given qt is not planning to release 5.8.x point releases, lists.qt-project.org/pipermail/releasing/2017-February/004471.html and relevant bugfixes should be manually backported b packagers.

bhush9:
Ping?

PR #1480, opened 2017-05-23
core/linux-armv7: Enable SPI_CADENCE_QUADSPI and JFFS2_FS

webmeister:
Those options are needed to access the flash memory on my Terasic SoCKit (Altera socfpga).


WarheadsSE:
It looks like you properly used the kernel's config scripts to update the config. Great.

I do have one question though: Should we enabled the `CONFIG_JFFS2_FS_XATTR`?

kmihelich:
Yes, XATTR should be enabled as well.  Also test whether or not these actually need to be built-in, since the kernel is already pretty fat.

webmeister:
I've enabled XATTR and switched both options to modules. This reduces the zImage size by ~40k.

PR #1484, opened 2017-05-27
alarm/kodi-rbp to 17.3-2

graysky2:
~~Note that this PR is dependent on https://github.com/archlinuxarm/PKGBUILDs/pull/1483~~

EDIT: It should be noted that the hardening-wrapper currently does not support distcc builds.  On my RPi3, it takes over 80 minutes to build this package without distcc compared to about 20 min with distcc.  If no distcc is a deal breaker, let me know and I will adjust the PR to not use the script.


kmihelich:
All builds use distcc.  If hardening-wrapper is included then it will just be bypassed.

graysky2:
Would you prefer that I remove it from the PKGBUILD?  This commit does also include the python-path fix.

kmihelich:
It can be removed, it's not going to be doing anything.

graysky2:
Done.  I have also closed PR#1483 is this is not needed.

PR #1491, opened 2017-06-22
alarm: create dkms-mali-utgard-meson

jakogut:
DKMS package for Mali Utgard kernel driver on meson platform


kmihelich:
Missing buildarch, and the note about kernel version in the package description is probably overkill.

jakogut:
Thanks for the feedback, updated.

kmihelich:
I thought this needed at least 4.12?  You have greater than in the depends.

jakogut:
Good catch, fixed. I also changed the pkgver to match the userspace mali libGL packages.

PR #1511, opened 2017-09-06
core/linux: Unify TPM support in kernel configurations

webmeister:
Arch Linux ARM runs on many platforms, but the state of TPM support is
different for most of the kernel builds. From a user's point of view, using
for example a Raspberry Pi 1 is not really different from using a Raspberry
Pi 3, so it would be nice to have the same level of TPM support across all
kernels, so that one can simply say that "Arch Linux ARM can use TPMs out
of the box", without needing to think too much about all the different
platforms.

Platform support:
* Enable CONFIG_TCG_TPM and everything below for Raspberry Pi and generic
  kernels.
* Enable CONFIG_HW_RANDOM_TPM for all kernels with TPM support, so that the
  TPM's random number generator can be used to get additional entropy.
* Build all TPM drivers as modules, which should be sufficient for current
  use cases and does not impact users without a TPM device.
* Do not change the configuration for Chromebook kernels, since those
  devices do have a TPM integrated, so its specific driver can be builtin.
* Do not enable TPM support for other platform-specific kernels that have
  it disabled at the moment, since it is not clear whether a TPM can easily
  be added to such a platform.

TPM support:
* Disable CONFIG_TCG_ATMEL everywhere, since this is a driver for very old
  devices that are most likely not used with any ARM-based machines.
* Enable (if supported) CONFIG_TCG_TIS_SPI, CONFIG_TCG_TIS_I2C_ATMEL and
  CONFIG_TCG_TIS_I2C_INFINEON for all kernels with TPM support, since those
  should cover the most common devices.

I've tested CONFIG_TCG_TIS_SPI and CONFIG_TCG_TIS_I2C_INFINEON with my collection of Raspberry Pis. I do not have an Atmel TPM to test CONFIG_TCG_TIS_I2C_ATMEL, but since this was already enabled for some platforms, it probably works for the others as well.


webmeister:
@kmihelich: Did you close this request intentionally without a comment?

webmeister:
@kmihelich: A little more communication from you would be appreciated. Or do you really want me to guess what you did not like about the request? :)

pepedog:
One package per request.
Also did you use make menuconfig or equivalent, and build packages?

webmeister:
Thanks for your reply. I had seen that rule, but since the whole point of this change was to ensure identical configurations for multiple packages, I assumed that grouping those changes in a single request was the better solution. But if you prefer 13 individual pull requests, I can accommodate that.

PR #1513, opened 2017-09-09
Add the install variable to kodi-c2 PKGBUILD

marcoscarpetta:
I think it was removed by mistake in 388d8bdadba05a9932c46ab924a84e3c87df4209.


marcoscarpetta:
The build failed, but the error is not related to this commit.
https://archlinuxarm.org/builder/in-log/kodi-c2-17.4-2-armv8.log.html.gz
The problem is that libxslt includes xlocale.h, which was removed from glibc in version 2.26 (https://github.com/bminor/glibc/blob/96480903ab6f16cba475c02e32827fa4bf2b62b7/NEWS#L191).
In the mail I was said to "review the build log and submit corrections for the package", but the corrections are to be done on the libxslt package.

kmihelich:
There's a new libxslt package, and this has built fine now.

PR #1516, opened 2017-09-20
core/linux-raspberrypi: re-enable BCM_VC_SM

invisiblek:
* This has been fixed in the raspberry pi kernel for multi-platform
  builds.
* Also fix building with a non-clean tree as firmware/brcm
  already exists in that case.


invisiblek:
Fixed here: https://github.com/raspberrypi/linux/pull/2205

invisiblek:
err, actually. i need to fix the hash for the kernel tar it grabs, hold off a sec

kmihelich:
No, the referenced PR does not fix the issue.  ARMv6 will still be running ARMv7 assembly.

invisiblek:
Huh? It's ifdef'd for v6/v7, unless I'm missing something. The referenced PR fixes the compile error, which I assumed is what caused you to remove this config option in the first place.

invisiblek:
Oh nevermind, I see what you're saying. I have a patch in mind.

PR #1520, opened 2017-09-26
core/linux-raspberrypi: re-enable BCM_VC_SM

invisiblek:
* This has been fixed in the raspberry pi kernel for multi-platform
* Also fix building with a non-clean tree (firmware/brcm already exists)


invisiblek:
Let's try this again ;)

kmihelich:
You've tested this on both architectures?

invisiblek:
I've only tested on ARMv7 (pi3), but it'll now run the ARMv6 assembly on v6 devices, which it hadn't been doing. I do have an original model B somewhere around that's not set up, but could test on that sometime later this week.

kmihelich:
I'd like to see this tested on both to be absolutely sure.

invisiblek:
Well, it builds, boots, /dev/vcsm exists, so its loading the driver correctly (ARMv6). But it doesn't appear the ARMv6 pi is affected by not having this driver in kodi. Kodi will crash on the pi3 if this driver isn't loaded, but not the case on the 1st gen. I suppose I'd have to dig into the userspace vc_sm stuff to know why.

Regardless, this should be good to go. The driver is now arch-agnostic upstream.

jonnybarnes:
Any idea when this might make it into the repos? I believe this will fix Kodi video playback by bringing back `/dev/vcsm`

graysky2:
It looks as though 4.9.52 has this present.  I building for v7 and v6 now.  I cannot test v6 on hardware though...

PR #1534, opened 2017-10-29
Add retroarch-rbp

faergeek:
Sorry, not sure on which architectures it will actually work, I have only raspberry pi 2.

Differences from upstream (also listed in PKGBUILD itself):
1. Add paths to`/opt/vc` to use videocore gl libs instead of mesa. This should and seems to be more performant.
2. Enable `neon` and `floathard` flags. Also for performance.


kmihelich:
https://aur.archlinux.org/

PR #1546, opened 2017-12-01
core/linux-am33x: add mcp23s08 and tps65217 backlight drivers

mniestroj:
These drivers are required to get use of chiliboard LCD
shield. mcp23s08 is a I2C GPIO expander needed to enable LCD.

Also update LCD shield device-tree to put LCD_EN pin high.


kmihelich:
Squash the commits, and update pkgrel and checksums in the PKGBUILD.

mniestroj:
Thanks for comments. Updated as requested.

PR #1557, opened 2018-01-28
Enable gcc-ada

vicencb:
Hello,
this pull request is bumping gcc version to 7.3.0 and enables gcc-ada.
gcc-ada is written in Ada, so, it requires bootstrapping.
I have successfully built and used it on aarch64 with the gnat compiler from Debian.

Regards,
  Vicente.


kmihelich:
Upstream Arch has not released this version, and merging this doesn't magically reproduce what you did.

vicencb:
I know that it will not automatically build as is, it requires bootstrapping which is a manual process.
In the patch itself there is a way on how to do the bootstrapping on aarch64, other architectures would follow a similar process.
Please, can you enable gcc-ada by honoring this PR or by other means?

Regarding the version: upstream has it in testing, I am fine waiting until it is in stable.

Regards,
&nbsp;&nbsp;Vicente.

PR #1568, opened 2018-03-10
add fmt 4.1.0-3 now a makedepends for kodi-rbp-git

graysky2:
This is a makedepends for the updated kodi-rbp-git package in https://github.com/archlinuxarm/PKGBUILDs/pull/1569


graysky2:
No needed.

PR #1569, opened 2018-03-10
kodi-rbp-git to 18.0a1.20180310-1

graysky2:


kmihelich:
Why not build with -DENABLE_INTERNAL_FMT=ON?

graysky2:
> Why not build with -DENABLE_INTERNAL_FMT=ON?

Good suggestion.  Let me try this and see if it works.

graysky2:
Good call.  Compiles for both successfully.  Rebased.

PR #1582, opened 2018-04-29
add odroid-xu3-clinfo package

mkaczanowski:
Customized version of clinfo community package that works with mali drivers used in odroid xu3/xu4.


kmihelich:
You can submit this to the AUR, which you seem ready to do anyway with the unnecessary .SRCINFO file.

PR #1583, opened 2018-05-05
wiringpi's /usr/bin/gpio lost a suid bit in a recent update

jernst:
It used to be `chmod 4755` in the `PKGBUILD`. Then the `PKGBUILD` was reworked, and the bit got lost, so `gpio` refuses to run.


jernst:
Fastest merge ever :-) Thank you.

PR #1584, opened 2018-05-07
[Add] Blender Package

initrider:
I've added blender package with some modifications from upstream


initrider:
@kmihelich what's wrong?

PR #1595, opened 2018-06-13
alarm/raspberrypi-firmware to 20180607-1

graysky2:
PR will bring firmware package in sync with kernel version and upstream.


kmihelich:
Still in sync as is.

graysky2:
I see now after comparing the md5sums of the two packages (after sorting them).  How can you tell before packaging and doing this comparison if any of the package-provided files have in fact changed?

PR #1596, opened 2018-06-14
xf86-video-armsoc-rockchip: Initialize RandR before ARMSOCEnterVT

ryan-farley:
This fixes a segfault caused by the use of RandR CRTCs in Xorg 1.20
before the driver had initialized them.


ryan-farley:
Thanks for fixing this.

It seems to apply to the other `xf86-video-armsoc` drivers as well (someone's XU4 on #archlinux-arm was fixed in a similar manner) but I'm not completely sure without the hardware to test myself.

PR #1610, opened 2018-08-12
community/mpv: fix the waf version

fmang:
mpv 0.29.0 requires waf 2.0.9, as specified by https://github.com/mpv-player/mpv/blob/v0.29.0/bootstrap.py#L8

The old version of waf makes the build break when Python is updated: https://github.com/mpv-player/mpv/issues/5958

The official PKGBUILD for mpv uses the right version already:
https://git.archlinux.org/svntogit/community.git/tree/trunk/PKGBUILD?h=packages/mpv#n10

### DIFF
--- a/community/mpv/PKGBUILD
+++ b/community/mpv/PKGBUILD
@@ -10,8 +10,8 @@
 pkgname=mpv
 epoch=1
 pkgver=0.29.0
-pkgrel=1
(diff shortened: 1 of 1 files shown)


kmihelich:
It's already built here.  When this gets bumped upstream, that change will be incorporated.

PR #1615, opened 2018-08-27
linux-{armv7,veyron}: add FIT image for mainline u-boot

hanetzer:
Personally I'm doing some work on u-boot to get the asus c201 chromebook
(veyron speedy) to work with u-boot. Mostly working, waiting on some issues
with 4GiB memory to be resolved.

U-boot can boot what's called a fit image which is built during the normal build
of linux-armv7-chromebook, just before the vbutil_kernel step. Install it to /boot
so it can be used by people using my u-boot patches, or later mainline u-boot

Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>

### DIFF
--- a/core/linux-armv7-rc/PKGBUILD
+++ b/core/linux-armv7-rc/PKGBUILD
@@ -386,6 +386,9 @@ _package-chromebook() {
     --bootloader bootloader.bin
 
   mkdir -p "${pkgdir}/boot"
+  # u-boot which supports FIT images
+  cp vmlinux.uimg "${pkgdir}/boot"
+  # standard chromeos coreboot+depthcharge
   cp vmlinux.kpart "${pkgdir}/boot"
 }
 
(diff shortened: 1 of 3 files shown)


hanetzer:
Pending an actual boot test, this should work, but building linux-armv7 on the c201 will
take quite some time due to low power in general.

kmihelich:
Mainline U-Boot works just fine with a zImage and dtb; there is no need for a fit image.

hanetzer:
@kmihelich true, but we make one anyways and its not a large file, considering the default
instructions you guys give on installation is one partition for depthcharge's ''boot'' partition
and the rest for /.

kmihelich:
...which works perfectly fine with U-Boot and all of the files already available in the linux-armv7 package.

hanetzer:
Oh, I was misreading what these packages installed. I thought it was just vmlinux.kpart.
Carry on :+1:

PR #1617, opened 2018-08-29
alarm: dkms-mali-panfrost-rk: add dkms package for panfrost mali

hanetzer:
Fairly simple PKGBUILD. Unsure about what to set the pkgver to since its from git
anyways. The dkms.conf is setup to build for rockchip, but it probably could be used
for other platforms.

Signed-off-by: Marty E. Plummer <hanetzer@startmail.com>


kmihelich:
Which kernel is this for?

hanetzer:
@kmihelich linux-armv7 4.18.1-1 is what I've build-tested against.

kmihelich:
How is this going to be used?

hanetzer:
Theoretically to be able to get mali gfx accell to work using linux-armv7 instead of linux-veyron,
but I've yet to find an actual guide/info on how to get it to work with linux-veyron/veyron-libgl.
The module inserts properly, and some simple tests work with it, but the actual 'using it' is escaping
me. Not that surprising, proprietary gfx are a pain quite a lot of the time.

kmihelich:
So there is no practical use.  Why does this need to be in the repository?

hanetzer:
There is, just as there is practical use for veyron-libgl. Its just not well documented on how to
actually get it going.

PR #1622, opened 2018-09-05
uboot-sunxi version bump, add orangepi zero and r1

sknsean:
Tested on Orange Pi Zero and R1


kmihelich:
Not supporting those boards here.

sknsean:
So how about we did? This could be a small step, they are very similar to the other sunxi boards...
I can do the setup guides and so on...

Will you take the version bump as an separate PR?

PR #1626, opened 2018-09-29
core/linux-armv7: add snd-dummy, snd-aloop and snd-virmidi

RoEdAl:


RoEdAl:
Now generated by `make menuconfig` command.

PR #1631, opened 2018-10-16
aur/cower: Added armv7h and aarch64

gounthar:
Built for aarch64 on my Raspberry Pi 3B+.


WarheadsSE:
@gounthar While this PR is appreciated, it is actually not needed.

[PlugBuild](https://github.com/archlinuxarm/plugbuild/) passes the  `-A` flag to `makechrootpkg` (see [this line](https://github.com/archlinuxarm/plugbuild/blob/master/client.pl#L355)), and as such, there is no need to modify the `arch=` array. We already have this built into, and available, for all provided architectures.

gounthar:
Great, thanks, but this package was not available when I tried to install it.
I must have missed something...

PR #1632, opened 2018-10-22
Add Sinovoip PI M2 Plus support

misbahxa:
I have successfully build and tested it.


kmihelich:
Not interested in supporting this device.

PR #1642, opened 2018-11-30
kodi-rbp-git needs libnfs as dependency

solsticedhiver:
So it seems kodi-rbp-git needs libnfs as dependency otherwise, kodi bails out at start with this error:

    /usr/lib/kodi/kodi-rbpi: error while loading shared libraries: libnfs.so.12: cannot open shared object: No such file or directory

This is a quick fix to let you know of the problem.
I haven't looked at how to do it with cmake variable

RANT:
Why can't we simply create issue and you deal with the problem ? Instead I have to create a pull request just to notify of the problem, and that pull request will never be accepted anyway ? wtf !


kmihelich:

PR #1647, opened 2018-12-25
core/linux-raspberrypi to 4.14.90-2

graysky2:
Correct me if I am wrong but I don't believe mkinitcpio is needed.  My test box boots and runs just fine without it.


kmihelich:
This is needed for the initramfs.  Just because you aren't using it doesn't mean no one else is.

PR #1650, opened 2018-12-30
Enabling USB3 on ROC-RK3328-CC (Renegade) like on Rock64

JpegXguy:
I don't know if it's missing regulator stuff that needs to be added, but it does work.


JpegXguy:
As can be seen [here](https://patchwork.kernel.org/patch/10179473/) it was suggested for merging into upstream but not ultimately merged because the replugging doesn't work. That probably also happens for the rock64 but it is included in the patches for archlinuxarm. I think it'd be fine if you included it temporarily until everything is fixed upstream. In fact the only patch needed is exactly this file (the original) sent upstream, also found in armbian's build.:

https://raw.githubusercontent.com/armbian/build/6a03dd4c0f05981fa6bcef1fd10429c2eaf5dc11/patch/kernel/rockchip64-dev/renegade-enable-usb3.patch

PR #1655, opened 2019-01-20
core/pacman to 5.1.2-3

graysky2:
If I'm not mistaken, we can drop `-fstack-protector-strong` since it's enabled by default in gcc.  Is this also true for Arch ARM toolchain?

References:
* [FS#54736](https://bugs.archlinux.org/task/54736)
* [Arch commit](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/pacman&id=c81d3556e03d4b3da79779a4804ba8bf8b127cb3)

### DIFF
--- a/core/pacman/PKGBUILD
+++ b/core/pacman/PKGBUILD
@@ -11,7 +11,7 @@
 
 pkgname=pacman
 pkgver=5.1.2
-pkgrel=2
+pkgrel=3
 pkgdesc="A library-based package manager with dependency support"
 arch=('x86_64')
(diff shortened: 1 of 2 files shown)


graysky2:
@kmihelich - I am incorrect that the ARM toolchain incorporates this by default as does the x86_64 one?

kmihelich:
1. It doesn't matter at all, I'll change it in the future.
2. https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md

PR #1659, opened 2019-01-22
Update linux-armv7-chromebook.install

quarkscript:
Each time when I update Arch Linux Arm on my chromebook (xe303c12) I need to flash a kernel manually, because kernel package can't find a partition to flash it. I think this mod can resolve this problem.

### DIFF
--- a/core/linux-armv7/linux-armv7-chromebook.install
+++ b/core/linux-armv7/linux-armv7-chromebook.install
@@ -1,8 +1,32 @@
+detect_kernel_partition() {
+  if [ -s "/boot/vmlinux.kpart" ]; then
+    hashalgo="sha256"
+    krnlsize=$(stat -c%s "/boot/vmlinux.kpart")
+    krnlhash=$("$hashalgo"sum /boot/vmlinux.kpart | sed 's/ .*//g')
+    for i in $(echo $(cat /proc/partitions | sed 's/name//g' | sed 's/.* //g')); do
+      tst=$(mktemp /tmp/XXXXXXXX.tmp)
+      dd if=/dev/$i of=$tst bs=$krnlsize count=1 status=none
+      if [ "$("$hashalgo"sum $tst | sed 's/ .*//g')" == "$krnlhash" ]; then
+        echo found kernel partition - /dev/$i
+        echo "$i">/boot/krnlpart
+        rm -f $tst
+        break
+      fi
+      rm -f $tst
+    done
+  fi
+}
+
 flash_kernel() {
-  major=$(mountpoint -d / | cut -f 1 -d ':')
-  minor=$(mountpoint -d / | cut -f 2 -d ':')
-  device=$(cat /proc/partitions | awk {'if ($1 == "'${major}'" && $2 == "'${minor}'") print $4 '})
-  device="/dev/${device/%2/1}"
+  if [ -s "/boot/krnlpart" ]; then
+    device="/dev/$(cat /boot/krnlpart)"
+    rm -f /boot/krnlpart
+  else
+    major=$(mountpoint -d / | cut -f 1 -d ':')
+    minor=$(mountpoint -d / | cut -f 2 -d ':')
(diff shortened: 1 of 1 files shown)


quarkscript:
Heh, my mod isn't applied, flashing error still present:
`(16/25) upgrading linux-firmware                   [######################] 100%
(17/25) upgrading linux-armv7                      [######################] 100%
(18/25) upgrading linux-armv7-chromebook           [######################] 100%
A new kernel version needs to be flashed onto /dev/.
Do you want to do this now? [y|N]
y
dd: failed to open '/dev/': Is a directory`

amstan:
Yeah, I have been thinking of something similar.

Just a little maybe-user defined indirection in flash_kernel would go a long way.

What would be even cooler is to do the vbutil_kernel step at runtime in the .install so the user can even customize the cmdline, but that would preclude having a downloadable .kpart. Eh, idk if this is good ROI, who would need a different cmdline?
I guess one could have signed kernels this way (if users feel like fiddling with their firmware keys they could get the real chromebook verified boot behavior in normal mode and alarm).

quarkscript:
Hi! 

I forked kernel pkgbuild earlier and modified it a little. Now my .install script looks like this:

```
flash_kernel() {
  krnprts=""
  krnprtc=0
  for i in $(ls /dev | grep -x --regexp="mmcblk[0-9]" --regexp="sd[a-z]" --regexp="vd[a-z]"); do
    if $(cgpt show /dev/$i | grep Label | grep -q [Kk]ernel); then
      if $(echo $i | grep -q mmcblk); then
        krnprts+="$i""p$(cgpt show /dev/$i | grep Label | grep [Kk]ernel -m 1 | sed '"'"'s/  / /g'"'"' | sed '"'"'s/  / /g'"'"' | sed '"'"'s/ La.*//g'"'"' | sed '"'"'s/.* //g'"'"') "
      else
        krnprts+="$i$(cgpt show /dev/$i | grep Label | grep [Kk]ernel -m 1 | sed '"'"'s/  / /g'"'"' | sed '"'"'s/  / /g'"'"' | sed '"'"'s/ La.*//g'"'"' | sed '"'"'s/.* //g'"'"') "
      fi
      krnprtc=$(($krnprtc+1))
    fi
  done
  
  echo ""
  if [ "$krnprtc" -eq 0 ]; then
    echo "Chrome OS kernel partition did not detected." 
    echo "You must reflash kernel manually."
  elif [ "$krnprtc" -eq 1 ]; then
    echo "Finded ChromeOs kernel partition - $krnprts"
    echo "Do you want to flash a new kernel to it? (y/n)"
    read -r shouldwe
    if [[ $shouldwe =~ ^([yY][eE][sS]|[yY])$ ]]; then
      dd if=/boot/vmlinux.kpart of=/dev/$krnprts
      sync
    else
      echo "You may flash kernel manually like:"
      echo "dd if=/boot/vmlinux.kpart of=/dev/$krnprts"
    fi
  else
    echo "Finded more than one Chrome OS kernel partition."
    echo "You need to select next action
    "
    echo "0 for do not flash"
    numpart=0
    for i in $krnprts; do
      numpart=$(($numpart+1))
      echo "$numpart for flash $i partition"
    done
    echo ""
    echo "Wrong partition may lead to bootfail. Be aware!"
    read -r shouldwe
    if [ $shouldwe -gt 0 ]&&[ $shouldwe -le $numpart ]; then
      numpart=0
      for i in $krnprts; do
        numpart=$(($numpart+1))
        if [ $numpart -eq $shouldwe ]; then
          dd if=/boot/vmlinux.kpart of=/dev/$i
          sync
        fi
      done
    fi
  fi
}
post_install () {
  flash_kernel
}
post_upgrade() {
  flash_kernel
}
post_remove() {
  rm -f boot/initramfs-linux.img
}
```
It uses cgpt to locate the ChromeOs kernel partition and then prompts to flash the kernel into that partition. 

As for kernel boot cmdline .install script may ask the user which kernel to use, the default or the re-signed one.

PR #1663, opened 2019-01-29
alarm/rbp3: add cortex53 specific package

graysky2:
We currently build with [Cortex-A7 optimizations](https://github.com/archlinuxarm/PKGBUILDs/blob/master/alarm/kodi-rbp/PKGBUILD#L101-L103) in [alarm/kodi-rbp](https://github.com/archlinuxarm/PKGBUILDs/tree/master/alarm/kodi-rbp) which covers the RPi2 series of hardware.  Since [upstream PR#13638](https://github.com/xbmc/xbmc/pull/13638), and gcc version >=7.2.1, compile-time optimizations that are specific for the Cortex-A53 processor in the Raspberry Pi 3 Model B and B+ can be added to boost performance on this hardware.  This PR adds a new package that uses them.


kmihelich:
What are the exact performance differences to support adding yet another package? I need to see a detailed comparison.

graysky2:
I will conduct some comparisons logging CPU% playing back some x265 video clips using each package and report the results here.  Traveling now.

graysky2:
95@kmihelich - I think Kodi is predominantly used for video playback, so I wanted a quantitative and reproducible measure of that to compare the two packages: **kodi-rbp-18.0-1** (currently in the alarm repo) or **kodi-rbp3-18.0-1** (built from this PR).

* Test hardware was a RPi3B+ outputting 1080p/60Hz.
* Test video clip was a 22 min x265 encoded mkv file (video bitrate = 3.110 kb/s and AAC audio bitrate = 500 kb/s).

I felt the best thing would be to measure response time on CPU intensive operations such as fast forwarding (2x or 4x) a video.  Unfortunately, I didn't have a good way to quantitate that, although my perception is that the **kodi-rbp3-18.0-1** build seemed both smoother and snappier when fast forwarding and rewinding than the **kodi-rbp-18.0-1** build was.

After thinking a bit more about something relevant and that could be measured, I came up with measuring kodi's CPU draw on just normal playback for the two builds.  I measured the CPU% of just the kodi pid using a python script called [psrecord](https://aur.archlinux.org/packages/psrecord/) when I started playback of the video.  Did this 3 separate times for each package.  The results showed an average _decrease_ in CPU usage of 9-12% with **kodi-rbp3-18.0-1**.  I can post the raw data logged if you want to do some statistics with it.

In summary:
With kodi-rbp-18.0-1
* mean CPU% = 162.95
* SD = 4.98

With kodi-rbp3-18.0-1
* mean CPU% = 148.89
* SD = 3.75

There may be better and more relevant ways to compare the two builds, but this demonstrates an advantage of the compiler flags unique to the RPi3's CPU.  If it's not too much trouble to add the package, I vote to do it.

alarm/kodi-rbp3/PKGBUILD
```diff
@@ -0,0 +1,228 @@
+# Contributor graysky <graysky AT archlinux DOT org>
+# Contributor BlackIkeEagle < ike DOT devolder AT gmail DOT com >
+# Contributor Oleg Rakhmanov <oleg [at] archlinuxarm [dot] com>
+# Contributor tomasgroth at yahoo.dk
+# Contributor WarheadsSE <max@warheads.net>
+# Contributor Adrian Fedoreanu <adrian [dot] fedoreanu [at] gmail [dot] com>
+
+buildarch=20
```
kmihelich: This needs to be corrected.

graysky2:
> This needs to be corrected.

Sorry about that, overlooked.  Fixed now/rebased/pushed.

PR #1554, opened 2018-01-05
alarm/kodi-odroid to 17.6

frol:
Tested on my Odroid U3

P.S. I couldn't get CMake configs working with Mali-EGL just yet.

PR #1665, opened 2019-02-03
alarm/kodi-odroid to 17.6-2 (based on alarm/kodi-c2)

frol:
Just for the record, there has been a prior conversation about the changes: https://github.com/archlinuxarm/PKGBUILDs/pull/1554#issuecomment-458707581

/cc @heikojansen


frol:
@heikojansen The package is already available from the repo. Could you, please, give it a try? Note, I have integrated the `kodi.service` you referenced.

heikojansen:
Tested it last night. Installed without problems and starting via systemd worked fine, too. (On second thought I did not test if "systemd enable kodi" works and leaves us with a running Kodi after reboot. Maybe tonight.)
This time the binary was also linked to a mariadb lib and not a non-existent mysql client lib.

So far everything looks good to me, so let me thank you again @frol  for your effort!

luukvbaal:
I'm trying to run this on my odroid XU4 but having difficulties with xorg drm permissions.
Might I bother you for any suggestions over at the [forum](https://archlinuxarm.org/forum/viewtopic.php?f=60&t=13413).
Thanks for updating the package.

frol:
@luukvbaal What are the permissions on `/dev/dri/card0`? (It is what gets probed according to your `Xorg.0.log`)

luukvbaal:
```
ls -l /dev/dri/card0
crw-rw----+ 1 root video 226, 0 Feb  8 01:05 /dev/dri/card0
```

frol:
@luukvbaal I have run out of ideas... What was the previous version of Kodi you run? How did you run Kodi previously? Also, the image was intended for Odroid X, X2, U2, and U3 (mostly because `odroid-libgl` targets these devices, so I thought it would only work for these devices), so I am not even sure whether this image should work for you.

luukvbaal:
Oh I seem to remember reading on the forum that the XU3 and XU4 are software compatible as far as the packages on the arch repo. Anyhow I only recently got the device and have yet to succeed in running kodi. So my issue is more of a problem with running startx which fails with the same error regardless of whether it is started by the kodi service. I am able to run xserver by running LXDM so this confuses me even more. 
In LXDM however kodi starts but is throwing errors about a missing `/usr/bin/kodi-xrandr` and shows a black screen. It does start however since I can alt-tab to it but am then unable to do anything in the UI with either keyboard/mouse/remote.

frol:
It seems to be the case that XU3 and XU4 are software compatible, but X, X2, U2, and U3 has little to do with XU* devices.

luukvbaal:
Oh my bad, I thought this package was for XU3 aswell, I guess I got confused by the `(ODROID-X/X2/U2/U3)` naming. So you're saying that that there is no kodi package for the XU3/4 in the repo's? Hardkernel's ubuntu image does come packaged with ubuntu so the device is compatible.

frol:
@luukvbaal It seems like this is the case. I am maintaining this package for my own U3, so there was no package for X/X2/U2/U3 before I step up.

luukvbaal:
Okay I guess I'll look into building from source If I can get `startx` fixed.

heikojansen:
Maybe have a look here: https://wiki.archlinux.org/index.php/Xorg#Rootless_Xorg ?
Other than that: packages odroid-xu3-libgl-fb and odroid-xu3-libgl-x11 are installed (see https://archlinuxarm.org/packages and search for "xu3" for the packages)?

PR #1468, opened 2017-03-21
alarm/kodi-rbp: spdif/toslink passthrough fix for hifiberry-digi+ card

nornic:
alarm/kodi-rbp 

This is basically copy of a pull request #1376.
As a reminder:

> The HiFiBerry Digi+ is a S/PDIF/Toslink output board for the Raspberry Pi (rPi2 and rPi3).
> 
> For years (I found forum topics going back to 2014) kodi is not able to use it as pass-through device for audio output. What that means is, you are not able to get DTS or DolbyD signal in your amp.
> 
> It looks OpenELEC and Osmc already applied the patch, or rather workaround for that issue.
> More info here:
> https://support.hifiberry.com/hc/en-us/community/posts/201845791-Kodi-no-passthrough
> and here
> http://forum.kodi.tv/showthread.php?tid=261778
> 
> Kodi won't apply the patch as in their eyes the problem should be fixed in kernel or by Hifiberry. 
> I really doubt it will be fixed upstream in the near future (judging on the sheer amount of topics/issues which are open all over the web).
> 
> It's worth applying that patch to save time and headache to arm arch users.

It looks like patch was removed when kodi package was upgraded to Krypton (v17) release.

Could it be merged again, please?

PR #1667, opened 2019-02-15
alarm/kodi-rbp: spdif/toslink passthrough fix for hifiberry-digi+ card

nornic:
alarm/kodi-rbp

This is a copy of pull request #1376 for 17.0 (and #1468 for 17.1) to work with kodi-rbp 18.0.

As a reminder:

> The HiFiBerry Digi+ is a S/PDIF/Toslink output board for the Raspberry Pi (rPi2 and rPi3).
> For years (I found forum topics going back to 2014) kodi is not able to use it as pass-through device for audio output. What that means is, you are not able to get DTS or DolbyD signal in your amp.
> It looks OpenELEC and Osmc already applied the patch, or rather workaround for that issue.
> More info here:
> https://support.hifiberry.com/hc/en-us/community/posts/201845791-Kodi-no-passthrough
> and here
> http://forum.kodi.tv/showthread.php?tid=261778
> Kodi won't apply the patch as in their eyes the problem should be fixed in kernel or by Hifiberry.
> I really doubt it will be fixed upstream in the near future (judging on the sheer amount of topics/issues which are open all over the web).
> It's worth applying that patch to save time and headache to arm arch users.

It looks like patch was removed when kodi-rbp package was upgraded to 18.0 release.

Could it be merged again, please?

### DIFF
--- a/alarm/kodi-rbp/PKGBUILD
+++ b/alarm/kodi-rbp/PKGBUILD
@@ -12,7 +12,7 @@ _prefix=/usr
 pkgbase=kodi-rbp
 pkgname=('kodi-rbp' 'kodi-rbp-eventclients' 'kodi-rbp-tools-texturepacker' 'kodi-rbp-dev')
 pkgver=18.0
-pkgrel=1
+pkgrel=2
 _codename=Leia
 _tag="18.0-$_codename"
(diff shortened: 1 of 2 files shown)


kmihelich:
@graysky2 ?

PR #1668, opened 2019-02-16
alarm/kodi-rbp3 - spdif/toslink passthrough fix for hifiberry-digi+ card

nornic:
alarm/kodi-rbp3

Just noticed there's a kodi package specific for rbp3, so this PR is just a copy of #1667 for kodi-rbp.

As a reminder:
> The HiFiBerry Digi+ is a S/PDIF/Toslink output board for the Raspberry Pi (rPi2 and rPi3).
> For years (I found forum topics going back to 2014) kodi is not able to use it as pass-through device for audio output. What that means is, you are not able to get DTS or DolbyD signal in your amp.
> It looks OpenELEC and Osmc already applied the patch, or rather workaround for that issue.
> More info here:
> https://support.hifiberry.com/hc/en-us/community/posts/201845791-Kodi-no-passthrough
> and here
> http://forum.kodi.tv/showthread.php?tid=261778
> Kodi won't apply the patch as in their eyes the problem should be fixed in kernel or by Hifiberry.
> I really doubt it will be fixed upstream in the near future (judging on the sheer amount of topics/issues which are open all over the web).
> It's worth applying that patch to save time and headache to arm arch users.

### DIFF
--- a/alarm/kodi-rbp3/PKGBUILD
+++ b/alarm/kodi-rbp3/PKGBUILD
@@ -12,7 +12,7 @@ _prefix=/usr
 pkgbase=kodi-rbp3
 pkgname=('kodi-rbp3' 'kodi-rbp3-eventclients' 'kodi-rbp3-tools-texturepacker' 'kodi-rbp3-dev')
 pkgver=18.0
-pkgrel=1
+pkgrel=2
 _codename=Leia
 _tag="18.0-$_codename"
 _ffmpeg_version="4.0.3-$_codename-RC5"
@@ -50,6 +50,7 @@ source=("https://github.com/popcornmix/xbmc/archive/newclock5_$_tag.tar.gz"
   "http://mirrors.kodi.tv/build-deps/sources/crossguid-$_crossguid_version.tar.gz"
   "http://mirrors.kodi.tv/build-deps/sources/fstrcmp-$_fstrcmp_version.tar.gz"
   "http://mirrors.kodi.tv/build-deps/sources/flatbuffers-$_flatbuffers_version.tar.gz"
+  "hifiberry_digi.patch"
 )
 noextract=(
   "libdvdcss-$_libdvdcss_version.tar.gz"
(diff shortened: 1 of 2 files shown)


kmihelich:
@graysky2 ?

graysky2:
I cannot recall why I dropped that patch.  If @nornic can confirm no breakage to systems without the hifiberry hardware, should be good to merge.  I see two pending PRs for the affected packages.  I am building the kodi-rbp-git package now with the patch and will test it on a non-hifiberry RPi3 when ready/rebase my PR to update that package.  Stay tuned.

nornic:
Unfortunately, I have no means to test this on non-hifiberry rpi3 without cannibalising current h/w:)
The patch never had any adverse affects, it only "spoofs" the card type when hifiberry card is detected. Does nothing if the cards is not there. The change should be pretty safe.

I'll keep an eye on this thread, just in case more work is needed.

Thank you!

graysky2:
@kmihelich - I tested kodi-rbp-git including this patch and I see no issues with playback/functionality on a stock RPi3.  My opinion is safe to merge this and the other PR.  I also rebased the PR against kodi-rbp-git including the patch as well.

graysky2:
@nornic - Would you be willing try a test?  

1. Downgrade to kodi-rbp-18.0-1 (you can use [this](http://tardis.tiny-vps.com/aarm/packages/k/kodi-rbp/) archive to grab the package.  
2. Create `/usr/share/alsa/cards/HifiberryDigi.conf` using [this file](https://github.com/LibreELEC/LibreELEC.tv/blob/libreelec-8.2/projects/RPi2/filesystem/usr/share/alsa/cards/HifiberryDigi.conf) and see if audio works.

LE does not patch their kodi source and as I understand it, the Hifiberry works using this method.  Thanks.

nornic:
@graysky2 sound works but no passthrough, which means I can't get DolbyD or DTS output.
Passthrough output option is greyed out even with the conf file in.
Let me play with this for a bit.

IIRC I took original patch from either OE or LE repository.

nornic:
OK, partial success.
Found the other forum [thread](https://forum.kodi.tv/showthread.php?tid=342129) you were discussing this, and as I have kernel 4.19.27, the file you linked here, didn't work at all.
I had to use [RPi-WM8804.conf](https://raw.githubusercontent.com/LibreELEC/LibreELEC.tv/master/projects/RPi/filesystem/usr/share/alsa/cards/RPi-WM8804.conf) which was mention in the same thread but in comment [#4](https://forum.kodi.tv/showthread.php?tid=342129&pid=2835433#pid2835433). File name is important, which caught me by surprise.

Once I downgraded to 18.0-1, added RPi-... file and reboot, I could finally select the passthrough mode in Kodi settings. Wasn't grayed out anymore. So success...
However, even when this passthrough is setup to hifiberry-digi, my amplifier only picks up PCM signal.

To prove I'm not going mental, removed the conf file, upgraded back to 18.1-2  and reboot. Any movie I switched on, had the passthrough working correctly. I got DolbyD and DTS audio.

Perhaps there's something else I am missing here... Don't know much about configuring the alsa cards via those conf files...

Any suggestions?

graysky2:
Hmmm... maybe it's best to take this to the kodi forum to figure out.  I only pinged you here since you're the only person on the planet I know that has the hardware :)

nornic:
> I only pinged you here since you're the only person on the planet I know that has the hardware :)

lol, back in the day, hifiberry was the best way to get good quality audio via spdif or toslink from rPi ;D

I subscribed to the other thread already.
Completely get your point of not patching kodi if we can achieve the same through other means.

Thanks for doing this.

PR #1677, opened 2019-03-08
core/linux-aarch64: Fix the problem that Rock64 USB ports are broken

parly:
Rock64 USB ports are broken since linux-aarch64 4.20.0-1.
This dts patch can bring back USB again, but it might take a long time to be included in upstream kernel release.
So I would like to request to apply it to linux-aarch64 package.

Related links:
* https://archlinuxarm.org/forum/viewtopic.php?f=23&t=13321
* https://lists.infradead.org/pipermail/linux-rockchip/2019-March/023298.html
* https://patchwork.kernel.org/patch/10843309/


parly:
Thank you for merging this!

PR #1678, opened 2019-03-13
core/linux-aarch64: fix usb 3.0 for rk3328-roc-cc

nachobeez:
Enable the USB3 port on the rk3328-roc-cc board
IT is a very important feature of the board, and useful when it's used as a NAS for example

Links:
- http://lists.infradead.org/pipermail/linux-rockchip/2018-January/019032.html (patch that added the board)
It was picked up by Heiko for 4.17 with the exception of the usb3 port due to a replug issue.

- https://raw.githubusercontent.com/armbian/build/master/patch/kernel/rockchip64-dev/board-renegade-add-enable-usb3.patch (just the usb3 bits for the dts, used in rockchip64-dev in armbian)


kmihelich:
Skirting bans from your blatant, irreverent disregard for posted rules and requirements and then continuing that tradition has now ensured your permanent placement on the shitlist.

PR #1682, opened 2019-03-21
alarm/kodi-rbp - Removed not needed hifiberry patch

nornic:
It seems the kodi hack for hifiberry Digi+ card which was applied as part #1667 is no longer needed.

Thanks to @graysky2 more info can be found on the [kodi thread](https://forum.kodi.tv/showthread.php?tid=342262).
The same functionality can be achieved by adding [RPi-WM8804.conf](https://raw.githubusercontent.com/LibreELEC/LibreELEC.tv/master/projects/RPi/filesystem/usr/share/alsa/cards/RPi-WM8804.conf) to **/usr/share/alsa/cards** and reboot.


graysky2:
@nornic - I think we might want to adjust your PR adding the requisite file you linked so that this package actually provides it thus retaining the ability to use the hardware natively.  Thoughts?  Can you do that and rebase?

nornic:
Not sure we should add the file to kodi package. Better add it to alsa-lib, where it'll be beside all other card configs. What do you think?

graysky2:
@kmihelich - Thoughts?

kmihelich:
I added it to the firmware-raspberrypi package.

PR #1686, opened 2019-03-27
core/linux-armv7: support flashing to chromebook internal storage

terceiro:
The current code assumes that the installation is done to partition
number 2, and consequently that the kernel is in partition number 1. I
installed Arch on an ARM Chromebook by first resizing the ChromeOS state
partition, and them using partitions 6 and 7 from the internal MMC for
kernel and rootfs, respectively.

This change makes the code generic: if the root partition is on
/dev/xxx(N), then it will flash (or suggest the user flashes) the kernel
to /dev/xxx(N-1).


alexdewar:
I guess an alternative approach would be to add e.g. a config file at ``/etc/kernel-partition.conf``, containing ``KERNEL_PARTITION=/dev/mmcblk1p2`` or whatever and then this could be sourced by the hook.

PR #1691, opened 2019-04-19
alarm/uboot-rockpro64: New package, version v2019.04

adam900710:
Half upstream, with only one commit ahead.

Test with real board, SDcard and ethernet works. No eMMC module to test
yet.

---

Unfamiliar with the board wiki page, any idea how to craft that wiki page?


adam900710:
Upstream have incomplete memory controller initialization for that board.

Using this version uboot will lead to random SError for memory access.

I'll replace the package with legacy build, and add `-mainline` suffix for the experimental build.

PR #1698, opened 2019-05-05
enable xt_set for kubernetes networking

everflux:
Kubernetes container networking depends on the xt_set capability, see https://github.com/weaveworks/weave/issues/2617#issuecomment-282577989


everflux:
Upstream: https://github.com/hardkernel/linux/pull/374

PR #1703, opened 2019-05-21
core/nfs-utils: Fix mountd segfault

PlanetEater:
Add patch from linux-nfs@vger.kernel.org that fixes a mountd
segfault when there are netgroups or hostname wildcards in
/etc/exports.


kmihelich:
This patch has not been accepted, acked, or even commented on by upstream maintainers.  Not merging.

PlanetEater:
The fix has been merged into upstream version 2.4.1.

PR #1708, opened 2019-06-21
Added MVSAS kernel module to allow external disks attached to SATA interface for Espressobin boards.

kurik:
The current kernel does not have mvsas kernel module built. Without this module the kernel does not work with integrated PCI/SATA interface on Espressobin v5 boards. As such a disk attached to this SATA interface is not recognized by kernel and the operating system can not use it.

Ref: https://archlinuxarm.org/forum/viewtopic.php?f=15&t=13675


kurik:
Rules for contribution were not followed. Will come up with one more try.....

PR #1716, opened 2019-07-18
Split linux-aarch64 into separate kernel and module packages

alexdewar:
Currently the ``linux-aarch64-chromebook`` package depends on ``linux-aarch64``, which includes both the kernel modules, as well as ``Image`` and ``Image.gz``. ``linux-aarch64`` also includes a hook for ``mkinitcpio``, which isn't needed for Chromebooks as they do not use the generated initramfs. By splitting the package into a separate ``linux-aarch64-modules`` package, we can avoid generating an unnecessary initramfs on Chromebooks.

If you guys are happy with this change I can also fix up other similar packages, e.g. ``linux-armv7``.

### DIFF
--- a/core/linux-aarch64/PKGBUILD
+++ b/core/linux-aarch64/PKGBUILD
@@ -8,7 +8,7 @@ _srcname=linux-5.2
 _kernelname=${pkgbase#linux}
 _desc="AArch64 multi-platform"
 pkgver=5.2.1
-pkgrel=1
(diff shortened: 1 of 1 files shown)


kmihelich:
This is a minimal overhead that really isn't worth splitting out.

alexdewar:
It is slightly annoying building an initramfs that you're not going to use, but it's up to you.

PR #1719, opened 2019-07-23
add kodi-rbp4 18.3-1 (no hw acceleration)

graysky2:
This gbm kodi package for RPi4 can only use sw decoding but does work.  I was able to playback even x265 content with it, but it does require all of the CPU to decode the sample I played.  As you know, I am working with @popcornmix to adapt his https://github.com/popcornmix/xbmc/commits/leia_pi4 fork (enables MMAL decoding for RPi4) which is what LibreLE uses, but it's not ready yet.  This would provide users with RPi4 hardware an option in the meantime.

### DIFF
--- a/alarm/kodi-rbp4/kodi.install
+++ b/alarm/kodi-rbp4/kodi.install
@@ -1,0 +1,26 @@
+post_install() {
+  [[ $(type -p gtk-update-icon-cache) ]] && usr/bin/gtk-update-icon-cache -qtf usr/share/icons/hicolor
+  [[ $(type -p update-desktop-database) ]] && usr/bin/update-desktop-database -q usr/share/applications
+  getent group kodi > /dev/null || groupadd -r kodi
+  getent passwd kodi > /dev/null || useradd -r -m -d /var/lib/kodi -g kodi -s /usr/bin/nologin kodi
+  usermod -a -G kodi,audio,video,power,network,optical,storage,disk,tty,input kodi
+  mkdir -p var/lib/kodi
+  chown -R kodi:kodi var/lib/kodi
+
+  echo "****************************************************************"
+  echo "Make sure the following lines are added to /boot/config.txt"
+  echo "gpu_mem=320"
+  echo "dtoverlay=vc4-fkms-v3d"
+  echo "dtparam=audio=on"
+  echo "****************************************************************"
+}
+
+post_upgrade() {
+  post_install $1
+}
+
+post_remove() {
+  [[ $(type -p gtk-update-icon-cache) ]] && usr/bin/gtk-update-icon-cache -qtf usr/share/icons/hicolor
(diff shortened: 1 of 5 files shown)


alarm/kodi-rbp4/kodi.install
```diff
@@ -0,0 +1,26 @@
...
+  [[ $(type -p gtk-update-icon-cache) ]] && usr/bin/gtk-update-icon-cache -qtf usr/share/icons/hicolor
+  [[ $(type -p update-desktop-database) ]] && usr/bin/update-desktop-database -q usr/share/applications
+  getent group kodi > /dev/null || groupadd -r kodi
+  getent passwd kodi > /dev/null || useradd -r -m -d /var/lib/kodi -g kodi -s /usr/bin/nologin kodi
+  usermod -a -G kodi,audio,video,power,network,optical,storage,disk,tty,input kodi
+  mkdir -p var/lib/kodi
+  chown -R kodi:kodi var/lib/kodi
+
+  echo "****************************************************************"
+  echo "Make sure the following lines are added to /boot/config.txt"
```
graysky2: Per our email regarding filtering for this file, we could omit this and just include a separate file this package would provide as a drop-in but that would require modifying `/boot/config.txt` provided by linux-raspberrypi

What do you think?

graysky2:
@kmihelich - I finally got the MMAL accel working but the build of it is non-standard currently requiring the modification of some dirs under the buildroot's `/opt/vc/include` which I do not think can be done from within `makechrootpkg` at all.

I am trying to come up with a work around.  Strategy is to simply copy the contents of raspberrypi-firmware to $srcdir, and then modify that copy and finally to redirect cmake to use it rather than the native location.

I am hitting some roadblocks currently.  See progress [url=https://forum.kodi.tv/showthread.php?tid=345572&pid=2871271#pid2871271]here[/url].

1) Any ideas are welcomed.
2) We still need to figure out how to handle mods to `/boot/config.txt`
3) Up to you if you want to merge this PR or wait for a solution to the build and merge that one.

kmihelich:
The path issue is something I'd wholly blame on the kodi sources, not the packaging here.  Everything else has been building with the packaged paths for years, and I can't think of any reason he is getting screwy with the paths now.

I'm going to wait and see about the config.txt changes.  If adding the overlay proves to be something standard across the board then I'll add it in for the Pi 4.

graysky2:
@kmihelich - The fork I want to provide is maintained by just one of their devs and is unofficial (they are working towards hardware decoding via v4l but this particular fork backported a number of ffmpeg MMAL optimizations).  It's sort of like the newclock5 branch we build for the other kodi-rbp packages: allows for amazing performance via HW decoding of HEVC content.

I would really like to see it build with `makechrootpkg` so others can enjoy it too.  Will send a PR if that happens.

PR #1723, opened 2019-07-29
alarm/kodi-rbp*: Explicitly specify the provided version of the kodi packages

ljani:
Arch Linux wiki states the following: "The version that the package provides should be mentioned (pkgver and potentially the pkgrel), in case packages referencing the software require one. For instance, a modified qt package version 3.3.8, named qt-foobar, should use provides=('qt=3.3.8'); omitting the version number would cause the dependencies that require a specific version of qt to fail." Source: https://wiki.archlinux.org/index.php/PKGBUILD#provides

This commit fixes building for example [`kodi-addon-inputstream-adaptive`](https://aur.archlinux.org/packages/kodi-addon-inputstream-adaptive/) from AUR, because it explicitly requests [`kodi-dev>=18`](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=kodi-addon-inputstream-adaptive&id=3cba01f7d208b8dab47caee8f987020220411a47).

See also: https://archlinuxarm.org/forum/viewtopic.php?f=15&t=13810

Sorry, I did not test building all of the packages, only `rbp3`, as I do not have a distcc setup available.

### DIFF
--- a/alarm/kodi-rbp/PKGBUILD
+++ b/alarm/kodi-rbp/PKGBUILD
@@ -158,7 +158,7 @@ package_kodi-rbp() {
     'upower: Display battery level'
   )
   install='kodi.install'
-  provides=('xbmc' 'kodi')
+  provides=('xbmc' 'kodi=18.3')
   conflicts=('xbmc' 'kodi' 'arm-mem-git' 'shairplay-git')
   replaces=('xbmc-rbp-git')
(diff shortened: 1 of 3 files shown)


ljani:
@kmihelich: Care to comment?

Here's a version with bumped `pkgrel`s as I forgot to do that.

PR #1726, opened 2019-08-05
Enable MCP251X CAN module for aarch64 kernel.

de-vri-es:
This PR enables the MCP251X CAN module for the aarch64 kernels.

It's a popular device for embedded systems, I think it makes sense to have the module available. Most of the other kernel configs already have it enabled.

I was wondering, should I enable it for the remaining kernel configs in this PR too? Or is there some reason it's not enabled for the other kernel packages? In any case, I did build the aarch64 kernel with this config successfully.


de-vri-es:
I'm sorry if I misunderstand the way to contribute to the PKGBUILDs. I was going by the message here: https://archlinuxarm.org/forum/viewtopic.php?t=2675, which seems to suggest I should open a PR here if possible.

If the PR is rejected for not bumping the pkgrel, I wasn't sure if I should do that or let the maintainer do it on the next rebuild.

I would really appreciate some form of comment on why the PR is rejected. The request itself doesn't seem unreasonable to me, considering there are other kernel packages here that already enable this option.

If I failed to follow the correct procedure, could you point me to the documentation of the right procedure?

PR #1727, opened 2019-08-06
GHC 8.6.5 fixes

brprice:
This package currently does not build. The first commit should fix that (stop trying to apply a patch that has now landed upstream), and the second changes to a supported LLVM.

I have not tested this package in a clean chroot, as ghc requires ghc to build, but we then need both llvm50 (for old ghc) and llvm6 (for the ghc we wish to build), and these conflict. I don't know how to resolve this, but I expect this has been done before in commit 32e26f1087b1f2aecca72c2cbd7fe1ffd1e2e158 (upgrading llvm39 -> llvm50).

I have manually verified that ghc builds with these changes, by keeping ghc-8.6.1-1 installed from the repositories and forcibly replacing llvm50 with llvm6 before doing the build.

### DIFF
--- a/community/ghc/PKGBUILD
+++ b/community/ghc/PKGBUILD
@@ -18,7 +18,7 @@ shopt -s extglob
 pkgbase=ghc
 pkgname=(ghc-libs ghc ghc-static)
 pkgver=8.6.5
-pkgrel=1
+pkgrel=1.1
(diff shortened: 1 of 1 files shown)


kmihelich:
I'm aware of these changes, and it's still problematic to build.

PR #1738, opened 2019-10-02
core/linux-aarch64: Enable 32bit compatibility

jou:
Currently, `linux-aarch64` has `CONFIG_COMPAT` disabled which makes it impossible to run 32bit binaries on 64bit machines. This PR enables `CONFIG_COMPAT` and emulation of deprecated instructions. This would enable running 32bit binaries down to armv6, like [Debian's AArch64 kernel does](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=861384).


kmihelich:
Not interested in supporting this at this time.

PR #1740, opened 2019-10-12
alarm: add uboot-rockpi4

adam900710:
This PKGBUILD is pretty much the same as previously attemped
uboot-rockpro4.

The main difference is:
- Uboot now has upstream support for LPDDR4 initialization
  The support comes in v2019.10 tag.
- Add dependency needed to compile upstream uboot
  New makedep includes python (for BL31/BL32) and dtc
- Remove PHY MAC address specification in boot.txt
  Now kernel/dtb/uboot can all use the persistent MAC address, no need
  to specify manually.
- Update to latest rkbin commit

Tested on RockPi 4B, 4G model.

Some known bugs related that board:
- Slow boot time due to watchdog hanging
  It will take 120s to initialize the tty and ethernet.
  It looks like to be caused by a watchdog hanging, looks more like a
  kernel bug.

- No PCIE (M.2) functionality
  The upstream hasn't yet enabled that feature.

- No wireless support yet

Despite that, upstream support should be pretty OK.


kmihelich:
Not an officially supported platform.

adam900710:
> Not an officially supported platform.

So that's why I'm adding the support. Or what's the proper way to submit a new board support?

PR #1741, opened 2019-10-13
Add grub package for aarch64

anarsoul:
Majority of aarch64 boards use u-boot as firmware. Since modern u-boot supports subset of UEFI that is enough to boot grub (and linux) we can actually make generic image even more generic by adding grub into it.

Unfortunately u-boot doesn't support persistent UEFI vars, so grub has to be installed as on removable media, i.e.:

`grub-install --no-nvram --removable --efi-directory=/boot/efi`

With grub installed and configured I was able to boot handcrafted image that contains 2 gpt partitions - 1st ESP mounted into /boot/efi, 2nd rootfs) on two different boards that have u-boot flashed into SPI flash - La Frite and Pine64-LTS.

Since grub-mkconfig expects linux image to be named vmlinuz-* it's also necessary to add `vmlinuz-linux` symlink to `Image.gz` otherwise it won't pick up kernel from linux-aarch64 package.

In theory now it's possible to create installation media like for x86/x86_64 platforms that can be flashed onto USB and used to install archlinux on aarch64 platforms.

### DIFF
--- a/core/grub/0003-10_linux-detect-archlinux-initramfs.patch
+++ b/core/grub/0003-10_linux-detect-archlinux-initramfs.patch
@@ -1,0 +1,41 @@
+diff --git a/util/grub.d/10_linux.in b/util/grub.d/10_linux.in
+index f5d3e78..ef59c8c 100644
+--- a/util/grub.d/10_linux.in
++++ b/util/grub.d/10_linux.in
+@@ -83,6 +83,8 @@ linux_entry ()
+       case $type in
+ 	  recovery)
+ 	      title="$(gettext_printf "%s, with Linux %s (recovery mode)" "${os}" "${version}")" ;;
++	  fallback)
++	      title="$(gettext_printf "%s, with Linux %s (fallback initramfs)" "${os}" "${version}")" ;;
+ 	  *)
+ 	      title="$(gettext_printf "%s, with Linux %s" "${os}" "${version}")" ;;
+       esac
+@@ -186,7 +188,7 @@ while [ "x$list" != "x" ] ; do
+   basename=`basename $linux`
+   dirname=`dirname $linux`
+   rel_dirname=`make_system_path_relative_to_its_root $dirname`
+-  version=`echo $basename | sed -e "s,^[^0-9]*-,,g"`
++  version=`echo $basename | sed -e "s,vmlinuz-,,g"`
+   alt_version=`echo $version | sed -e "s,\.old$,,g"`
+   linux_root_device_thisversion="${LINUX_ROOT_DEVICE}"
+ 
+@@ -248,6 +250,18 @@ while [ "x$list" != "x" ] ; do
+ 
+   linux_entry "${OS}" "${version}" advanced \
+               "${GRUB_CMDLINE_LINUX} ${GRUB_CMDLINE_LINUX_DEFAULT}"
++
++  if test -e "${dirname}/initramfs-${version}-fallback.img" ; then
++    initrd="initramfs-${version}-fallback.img"
++
++    if test -n "${initrd}" ; then
++      gettext_printf "Found fallback initrd image(s) in %s:%s\n" "${dirname}" "${initrd_extra} ${initrd}" >&2
++    fi
++
++    linux_entry "${OS}" "${version}" fallback \
++                "${GRUB_CMDLINE_LINUX} ${GRUB_CMDLINE_LINUX_DEFAULT}"
++  fi
++
+   if [ "x${GRUB_DISABLE_RECOVERY}" != "xtrue" ]; then
+     linux_entry "${OS}" "${version}" recovery \
+                 "single ${GRUB_CMDLINE_LINUX}"
--- a/core/grub/0004-add-GRUB_COLOR_variables.patch
+++ b/core/grub/0004-add-GRUB_COLOR_variables.patch
@@ -1,0 +1,32 @@
+diff --git a/util/grub-mkconfig.in b/util/grub-mkconfig.in
+index 3390ba9..c416489 100644
+--- a/util/grub-mkconfig.in
++++ b/util/grub-mkconfig.in
+@@ -218,6 +218,8 @@ export GRUB_DEFAULT \
+   GRUB_THEME \
+   GRUB_GFXPAYLOAD_LINUX \
+   GRUB_DISABLE_OS_PROBER \
++  GRUB_COLOR_NORMAL \
++  GRUB_COLOR_HIGHLIGHT \
+   GRUB_INIT_TUNE \
+   GRUB_SAVEDEFAULT \
+   GRUB_ENABLE_CRYPTODISK \
+diff --git a/util/grub.d/00_header.in b/util/grub.d/00_header.in
+index d2e7252..8259f45 100644
+--- a/util/grub.d/00_header.in
++++ b/util/grub.d/00_header.in
+@@ -125,6 +125,14 @@ cat <<EOF
+ 
+ EOF
+ 
++if [ x$GRUB_COLOR_NORMAL != x ] && [ x$GRUB_COLOR_HIGHLIGHT != x ] ; then
++    cat << EOF
++set menu_color_normal=$GRUB_COLOR_NORMAL
++set menu_color_highlight=$GRUB_COLOR_HIGHLIGHT
++
++EOF
++fi
++
+ serial=0;
+ gfxterm=0;
+ for x in ${GRUB_TERMINAL_INPUT} ${GRUB_TERMINAL_OUTPUT}; do
--- a/core/grub/PKGBUILD
+++ b/core/grub/PKGBUILD
@@ -1,0 +1,306 @@
+# Maintainer : Christian Hesse <mail@eworm.de>
+# Maintainer : Ronald van Haren <ronald.archlinux.org>
+# Contributor: Tobias Powalowski <tpowa@archlinux.org>
+# Contributor: Keshav Amburay <(the ddoott ridikulus ddoott rat) (aatt) (gemmaeiil) (ddoott) (ccoomm)>
+
+# ALARM: Vasily Khoruzhick <anarsoul@gmail.com>
+#  - Removed call to _build_grub-common_and_bios() since
+#    it's not supported on aarch64
+#  - Drop everything but packaging /etc/grub/default in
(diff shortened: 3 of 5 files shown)


anarsoul:
Any feedback?

anarsoul:
@kmihelich could I at least get some explanation why it's closed? You'd have to support EBBR [1] eventually and grub is a step forward in this direction.

[1] https://github.com/ARM-software/ebbr

superna9999:
@kmihelich hi, this is actually a very serious PR
I’m maintainer of the Amlogic Upstream U-Boot port and various Linux bits, and the subject of generic boot image using Grub via UEFI on arm/arm64 is serious and part of the ARM initiated Embedded Base Boot Requirements.
@anarsoul pointed 2 upstream supported boards, but this feature would make the generic image bootable on virtually all upstream supported boards with u-boot already flashed on the board.

Could this PR be re-opened and discussed seriously ?

kmihelich:
Sure.

1. The rules for contribution have not been followed.  A complete explanation of why is necessary.  These rules are not at all unique to this project, are not new, and are in place to enforce rigor in contributions, particularly from first time contributors such as @anarsoul.  Disrespecting the rules is very much not appreciated.  There exists a multitude of past commits and no-questions-asked PR merges to serve as examples of how to structure your PR.
2. @anarsoul has removed all upstream credit from the actual maintainers of the grub package, and has installed himself as the maintainer.  As a first time contributor and knowing the history of first time contributors, I have nearly zero expectation and absolutely no proof that this is anything beyond a drive-by PR.  Thus, knowing that I will be merging and applying changes from the upstream maintainers going forward, and that I have immense respect for their free time commitments and contributions to the distribution that makes this one possible, the stripping of the rightful attribution is incredibly poor form and was the top-most reason for outright dismissing this PR.  This point will also receive a complete explanation.
3. Given that upstream changes will be merged to keep this package in step, the packages in this repo should have the absolute minimum differences necessary from upstream to ease this process.  After reviewing the diff, I'm not convinced that this is the case in this PR.
4. Explain in all necessary technical detail why a vmlinuz symlink is required, and why grub can't be pointed to the existing Image or Image.gz files directly.

More information on which manufacturers have agreed to support the EBBR going forward would also be good to see.  Traceable statements of such, relevant manufacturer repos, or press releases are things that come to mind.  I see the potential for this in having a distribution image that doesn't require me to additionally maintain a U-Boot build.  However, if manufacturers don't adopt this and I'm still maintaining a U-Boot build, I'm not seeing a difference between configuring the build for the typical distro-boot scripting versus UEFI to chain-load grub, which creates another point of failure in the boot process.

anarsoul:
Thanks for response!

I believe I followed the rules from https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md

- PR title is descriptive
- Detailed description is here
- I can't do one package per request since adding grub requires a change in linux-aarch64 package
- Commits for individual packages are squashed
- My changes build
- That's the only PR I opened
- I detailed changes to upstream package in the comment header at the top
- Package is not yet in alarm repo => no need to increase pkgrel
- Package does not exceed version upstream
- buildarch is set to build it only for aarch64

As for credits - I did remove contributors, my fault, I'll re-add them. I'm not sure what to do with maintainers though. Should I preserve upstream maintainers list? Is it some way to put your copyright in PKGBUILDs? I put my name just to indicate that I'm willing to maintain this package for alarm.

I did maintain the minimal possible difference for this package. I dropped grub-bios part since it can't be built for aarch64 and moved /etc/grub/default into grub-efi-aarch64. Any suggestions how to minimize these changes are welcome.

As for symlink -- that's the way grub configs are written. grub-mkconfig scans all available images and initrds to generate an entry in menu for each but they have to be named in specific way. Archlinux for x86_64 has a suffix for them, i.e. core/linux installs vmlinuz-linux and mkinitcpio generates initramfs-linux.img.

I can change grub configs to pick up Image but that will result in even more changes to upstream package.

I'm not sure where to find information on adopters of EBBR, but keep in mind that upstream u-boot already supports it. So virtually any board supported by upstream u-boot supports EBBR. I'd still recommend to reserve some space for bootloader (just make sure first partition starts at sector 2048) if you're planning to supply ISO (or VFAT) image for boards that don't have SPI flash.

kmihelich:
There is no existing dependency between any kernel package and a currently non-existent grub package.  Adding a grub package also does not absolutely require any kernel changes, as any kernel packaging changes would be to further the usefulness of the grub package, and do not create any explicit hard dependencies between them in terms of build order or ability to build at all.  Your reasoning for ignoring the separation of packages between pull requests for specific review and discussion is simply not valid.

You have not sufficiently explained why you decided that the upstream authors did not deserve credit for their past and continuing work.  You are submitting a change to an upstream package for suitability and ability to build on ARM, not a separate work to be maintained wholly independently of the upstream package maintenance and thereby arguably necessitating a removal of upstream persons to alleviate a potential source of confusion for users or to distinctly separate ownership.  A serious discussion was requested, and I want a serious answer to this.

There is no reason to remove the x86 components or logic from the PKGBUILD if it is not necessary and not directly impacting the ability to build for ARM.  This goes to the point I already made about making my job easier maintaining this moving forward as upstream Arch revs the package and changes need to be merged.

You also haven't adequately responded to the implied seriousness of adding this package, in my opinion.  The argument as I see it is that U-Boot is to be shipped as a sort of UEFI BIOS by manufacturers, so seeing evidence of that support by manufacturers is key.  You implied that this is going to be the norm in the future, so you should be able to substantiate this further than two boards and a spec from ARM.  If the argument is instead that distributions continue to generate the necessary U-Boot image and comply with a standard that manufacturers don't care about, then layering on grub as an additional point of failure is not particularly appealing.  I would be far more in favor of continuing with distro-boot scripts for a simpler, more easily debugged boot process.  This is not to say that a grub package does not have a place here; I'm asking for my education and full understanding of the push toward this methodology.

anarsoul:
Yet default grub configs require kernel image to be named in a specific way, so either package - grub or linux requires modification.

I believe upstream authors deserve credit for their contribution but I guess "maintainer" title confused me. Usually maintainer is a person who maintains the package, and this person is not necessarily the author of packaging scripts. If any issue arises people usually go ask maintainers, not authors. Since upstream maintainers likely are not willing to maintain the package for ARM I decided to omit their names. I'm sorry for this mistake and I'll fix it in next revision.

grub-bios can not be build for aarch64. Following won't work:
```
        echo "Run ./configure for bios build..."
        ./configure \
                --with-platform="pc" \
                --target="i386" \
                "${_EFIEMU}" \
                --enable-boot-time \
                "${_configure_options[@]}"

        echo "Run make for bios build..."
        make
```

So I don't understand why you keep pushing me to preserve it. Do you want me to comment it all out and keep it as empty package? That would be essentially the same maintenance burden.

As for support from manufacturers -- I don't think it requires any extra support from them. As long as board uses u-boot you're able to boot using UEFI. Pine64 (I'm not affiliated with it, I'm just a hobbyist) doesn't declare EBBR support for Pine64-LTS, yet with u-boot flashed into SPI flash I'm able to boot. I can do the same with RockPro64 (tried that, it works, but u-boot doesn't support video output for rk3399 yet) and probably Rock64 (haven't tried that yet).

Anyway, what do you want me to do to get this PR merged? Is this list complete?

1. Restore original list of maintainers and contributors in PKGBULD
2. Drop changes to linux-aarch64 and linux-aarch64-rc (note that grub-mkconfig won't create any menu entries in this case)
3. Preserve grub-bios, but comment everything in it out, so essentially keep it empty?

Let's try to be constructive to get things done.

After all I'm trying to help and I'm doing it in my spare time (I'm not affiliated with any board manufacturer)

kmihelich:
What I'm looking for are compelling reasons to chain-load grub.  If there is a new trend with manufacturers toward preloaded SPI flash with a UEFI-capable U-Boot, as seems to have been implied, then there should be proof of this.  I don't see a compelling reason to chain-load for the sake of chain-loading.  U-Boot is perfectly capable on its own to start an OS, and can provide a similar experience to grub via extlinux.conf for those that want that kind of environment.  There has been an urgency and inevitability expressed by you two; I'm trying to understand what is behind that.  This distribution is my free time project, and I can't keep up on everything that's happening in the ARM world all the time.  I feel like I'm missing something here.

I'm not disagreeing with adding a vmlinuz-linux symlink.  I asked why and you gave a good answer.  I am disagreeing with your assumption that grub cannot exist without it as an explanation for creating a pull request targeting three packages.

Regarding the grub package:

1. Setting _IA32_EFI_IN_ARCH_X64="1" does not need to be removed, as it has no impact on the modifications for ARM.
2. Setting _EFI_ARCH and _EMU_ARCH for x86 are likewise inconsequential, and can remain.
3. Changing the arch array to aarch64 is meaningless.  All packages are built with -A to makepkg, and this change is in the most frequently updated area of a PKGBUILD and can cause patch failures.
4. _build_grub-common_and_bios() is an acceptable removal, but the same result can be achieved by commenting the function call.  This would then mean changes that upstream makes to this section apply smoothly, even if they do not impact the package here.
5. The call to _build_grub-efi for i386 is protected by a conditional that will be false for ARM, therefore it does not need to be removed either.
6. _package_grub-common_and_bios() could simply retain the last two lines that you moved.
7. 6 means that those moved lines are no longer adding a difference to _package_grub-efi().
8. As with 5, the call to _package_grub-efi for i386 is protected by a conditional that will be false for ARM, and does not need to be removed from package().
9. With retaining _package_grub-common_and_bios(), that call no longer needs to be removed from package().

For a modified package that is tracking upstream, a minimal diffstat matters more for the reduced maintenance overhead than the beauty of a package that doesn't have unused cruft for other architectures.

dsx724:
I am from Libre Computer (www.libre.computer) and we include proper u-boot implementation that supports the EBBR spec out of the box. We cannot speak for other manufacturers but every current board with SPI NOR will have an upstream u-boot implementation. We are also standardizing around it for all future boards.

anarsoul:
Sorry for delayed response, but I had no time to work on it last weekend.

I believe current version has all the issues addressed.

kmihelich:
Definitely in better shape.  You're a couple changes behind upstream now, though:
https://git.archlinux.org/svntogit/packages.git/log/trunk?h=packages/grub

anarsoul:
Now it's up to date with upstream

PR #1743, opened 2019-11-08
add linux-gameshell

r043v:
added an initial linux kernel image to support gameshell handled from clockworkpi

https://www.clockworkpi.com/gameshell

currently based on linux-armv7-rc package

current revision : 5.4.rc6


kmihelich:
Not a supported platform.

r043v:
yes, that's why i publish this
i currently got a full functional Arch on the platform,
result of some month of work to replace the stock Debian initially present
last thing keep closed is the constructor u-boot but i also work on this

what should i do to put my work in this official place ?
else does i need put it on classic AUR ?

PR #1749, opened 2019-12-16
Fix unapplied ARM linux kernel patches in PKGBUILD

abdel-unxp:
PKGBUILD is wrong for aarch64 linux kernel:
The linux source code is downloaded as tgz and then patches are applied
with git apply.
It is not allowed to do so, and doing so lead to unapplied patches
without throwing error. So lots of the patches are not applied.

This patch use patch comand instead of git aply.

Signed-off-by: Abdel Benamrouche <abdel@unxp.fr>

### DIFF
--- a/core/linux-aarch64/PKGBUILD
+++ b/core/linux-aarch64/PKGBUILD
@@ -47,14 +47,14 @@ prepare() {
   cd ${_srcname}
 
   # add upstream patch
-  git apply --whitespace=nowarn ../patch-${pkgver}
+  patch -Np1 -i ../patch-${pkgver}
 
   # ALARM patches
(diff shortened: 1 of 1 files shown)


rubenvb:
Why was this PR closed?

abdel-unxp:
No idea

WarheadsSE:

> Pull requests that fail to meet these requirements may be summarily closed without response.

Let me also give you some context on why this PR is not needed: You should not be building the packages from the Git tree itself. Copy the contents of that package to a working folder (we literally tarball the working directories and ship them to the Plugbuild clients), and run `makepkg` (better yet, `makechrootpkg`). You will have no issues with `git apply` then.

PR #1753, opened 2020-01-03
added alarm/uboot-orangepi-rk3399

jakogut:
This patch adds the uboot package for the Orange Pi RK3399 and Pi 4, based on the uboot-rock64 package. Tested on the Orange Pi 4.

The package works well using similar instructions to the Rock64. I can follow up with a wiki entry documenting the installation process for these new boards.

A patch is required to the rk3399-orangepi DTB to get the Pi 4 booting, I've sent an email to the relevant maintainers asking for advice on how to proceed, though we could also add it to the Arch Linux ARM packages if need be.


kmihelich:
Not a supported platform.

jakogut:
@kmihelich Is there a process for adding new supported platforms?

PR #1756, opened 2020-01-11
Add ovmf package for qemu aarch64

tomty89:


tomty89:
Anything I can improve to get this in? We have qemu and linux-aarch64 is built with KVM support. Having a UEFI for it avoids all the kernel and whatsoever extraction and/or qemu switches for booting.

tomty89:
Would you mind adding the package in the way that fulfill your standard if this doesn't?

tomty89:
Ugh that's pretty rude.

PR #1764, opened 2020-01-21
upgraded devtools-alarm to 20191227

petronny:
I've only tested it with [devtools-qemu](https://github.com/arch4edu/devtools-qemu).

### Important changes
I  moved
```
-C <dir>   Set pacman cache to pass to arch-nspawn
```
to
```
-X <dir>   Set pacman cache to pass to arch-nspawn
```
since `-C` is occupied by the official devtools now.

### DIFF
--- a/alarm/devtools-alarm/0001-makechrootpkg-cache-dir.patch
+++ b/alarm/devtools-alarm/0001-makechrootpkg-cache-dir.patch
@@ -15,7 +15,7 @@ index 2407115..f60d978 100644
  	echo 'Flags:'
  	echo '-h         This help'
  	echo '-c         Clean the chroot before building'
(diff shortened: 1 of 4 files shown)


kmihelich:
I'm handling this.

petronny:
Nice to hear that.

BTW, I've tested the zstd support.
zstd might raise OOM error with the default configurations.
You may need to adjust the number of cores in the `-T0` parameters.

There are 2 limitations for this.
The total RAM pre-allocated should be less than your RAM capacity.
And for 32-bit systems like `armv{6,7}h`, the total RAM pre-allocated should be also less than 4GB.
(Each core needs up to 400MB RAM with `--ultra -20`.)

petronny:
Hi, how is everything going?

PR #1765, opened 2020-01-25
[blender] Fixed ARMv6/7 building and enabled Cycles

satcom886:
The biggest thing here is that I enabled Cycles, and building for ARMv6 and ARMv7. However I couldn't test if this builds on ARMv6 because I keep running out of RAM on my Pi 0 W. But theoretically it should build.
I had to apply a patch for building on ARMv6 and 7 from [here](https://devtalk.blender.org/t/raspbian-blender-as-a-python-module-build-errors/10473/16).

### DIFF
--- a/community/blender/PKGBUILD
+++ b/community/blender/PKGBUILD
@@ -6,8 +6,18 @@
 #  - remove makedepend, optdepend on cuda
 #  - remove depend on openimagedenoise
 #  - set -DWITH_CYCLES=OFF, -DWITH_RAYOPTIMIZATION=OFF
+# ALARM: Rostislav Medvěd <rostik.medved@gmail.com>
+#  - Removed the unnecessary Optix source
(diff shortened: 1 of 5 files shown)


kmihelich:
32-bit is no longer tested or supported by Blender.

satcom886:
While 32 bit is not officially supported as they said [here](https://lists.blender.org/pipermail/bf-committers/2019-August/050124.html), they also mention what it means to "no longer be supported". "We don't actively test them (builds on those architectures) or release our own builds". I also think I know the reason why 32bit isn't supported any more. 2.81 introduced the Intel OpenImageDenoise feature, but OID requires a 64bit CPU with SSE4.1. Scince the ARM builds have OID disabled anyway, I don't see a reason not to merge this. From what I learned on the Blender dev forum the patches won't be needed in 2.82.

PR #1767, opened 2020-02-03
drop mkinitcpio and depmod hooks from linux packages

eworm-de:
The packages kmod (60-depmod.hook) and mkinitcpio
(90-mkinitcpio-install.hook) ship their own generic hook files. Do not
run twice but remove the package specific hooks.


eworm-de:
Oh, wait... This fails for some conditions. The initramfs is generated twice if trigger `usr/lib/initcpio/*` matches, but only once on `linux-*` package update. The linux packages do not ship `usr/lib/modules/*/vmlinuz`.

kmihelich:
Yeah, I haven't spent the time yet to think about how I want to solve this one.

eworm-de:
Well, I run archlinuxarm on one single Raspberry Pi (with `linux-raspberrypi`)...

I do not have a good overview, but for me it looks like...
- dropping the depmod hooks should be safe and...
- the generic mkinitcpio hook needs a matching trigger for all linux-* packages.

Is there a good reason (not?) to ship `/usr/lib/modules/*/vmlinuz`?

kmihelich:
I've considered adopting the Arch method of renaming the zImage or Image files to vmlinuz in modules, and letting that get installed as /boot/vmlinuz-linux.  However, this requires updating all of the bootloaders so it needs to be a more controlled roll out.  People will also likely miss or ignore the boot scripts being updated if they made local modifications, and end up with a system not booting.  This will be seen most from Pi users who usually have modifications, and a kernel= line will need to be added.  Unofficially supported systems with community bootloaders will similarly break when /boot/(z)Image disappears.

tillmann:
Why not ship with /usr/lib/modules/*/vmlinuz and install to /boot/Image (at least temporarily). Or even, let whichever boot method one wants to use deal with contents of /boot (say, sd-boot). Would seem a bit cleaner to me if the linux kernel package didnt interfere with the boot process. ( #1910 could be relevant)

I stumbled upon this while trying to install using pacstrap and seeing the mkinitcpio getting built way too often. Also, in the current state it is ugly to use dracut or booster for generating an initrd (which I wanted to do).

dvzrv:
@kmihelich is there any news on this? Currently everything still gets built twice. :cry: 

The fallback images are also still enabled by default (and can not be removed) - this makes four builds for only one kernel (of which 3 are unnecessary). It would be really great to solve this.

If you need help with it, please let me know!

tleydxdy:
I feel like breaking changes are to be expected in arch if there's a significant problem that needs to be solved (new archlinux arm user here so no idea if this is true for archlinux arm). Building the initramfs twice seems like a significant problem, it takes quite some while on my boards at least and significant % of the compute. So I would really wish this can be solved.

If breaking changes need to be avoided, maybe some kludge could work. e.g. the new kernel package with the fix is introduced with a different package name and conflicts with the old package so that new installs gets the fix automatically and old installs need to be manually switched over to the new package and the old package can be dropped from the mirrors. After awhile the new package can rename back to the nominal name.

Cornelicorn:
@graysky2 Why has this issue been closed?

PR #1769, opened 2020-02-23
Added extra/avidemux

jgmdev:
Changes to upstream:

- Changed build flags as shared here:
   https://avidemux.org/smif/index.php/topic,18851.msg87324.html
- Disabled cpu type check rule on a cmake file
- Added makedepends a52dec and libmad

### DIFF
--- a/extra/avidemux/PKGBUILD
+++ b/extra/avidemux/PKGBUILD
@@ -1,0 +1,85 @@
+# Maintainer: Levente Polyak <anthraxx[at]archlinux[dot]org>
+# Maintainer: Eric Bélanger <eric@archlinux.org>
+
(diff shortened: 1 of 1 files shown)


kmihelich:
Why are these changes needed?  32-bit ARM appears to build fine now as-is, and an upstream patch fixes AArch64.

jgmdev:
I took the official archlinux x86_64 PKGBUILD which is at version 2.7.4 and tried to build it on aarch64 but it failed complaining about not supported CPU architecture so I removed that CPU check, also the default bootstrap configure flags didn't worked and build failed so the flags needed to also be modified as found on the shared url. Finally since the build is linking against system liba52 and libmad I added those as makedepends. 

I haven't tested avidemux 2.7.5 which hasn't been promoted officially yet.

PR #1770, opened 2020-02-26
core/linux-odroid-n2 to 4.9.213 and fix for missing cursor

jgmdev:
Updated to 4.9.213 and enable console software cursor to fix missing cursor.

More details: https://forum.odroid.com/viewtopic.php?f=177&t=37431

The patch on https://forum.odroid.com/viewtopic.php?f=177&t=37431#p278056 for drivers/amlogic/media/osd/osd_fb.c isn't needed because it has been mainstreamed, only disabling CONFIG_AMLOGIC_MEDIA_FB_OSD2_CURSOR is needed now.


kmihelich:
Why are you hand editing the config?

jgmdev:
because I'm a noob when it comes to kernel configuration and compilation, not sure what are the proper steps to follow.

kmihelich:
Don't submit PRs if you don't know what you're doing, especially kernel packages.

jgmdev:
I compiled it, worked and cursor is now visible, but I will have to investigate how to properly modify the config file, but in this case no other dependencies need to get disabled...

PR #1777, opened 2020-03-12
core/linux-espressobin: build .dtb for espressobin-v7

greg2010:
In Espressobin v7 Globalscale swapped around ports lan1 and wan. This is not easily fixable without modifying the device tree file.
Fortunately, upstream has implemented a .dts for v7. Unfortunately, it is not built with flag `CONFIG_ARCH_MVEBU` (or any other flag afaik). This patch fixes that.


greg2010:
Would it also make sense to provide `alarm/uboot-espressobin-v7` with `fdt_name` targeting the new .dtb? If so, I can open a PR for that too.

greg2010:
Fixed upstream.

PR #1778, opened 2020-03-17
core/linux-odroid-c2 to 3.16.82-1

mdrjr:


kmihelich:
I have a feeling that dropping seccomp is going to break something, but I can't think of what that is right now..

PR #1779, opened 2020-03-20
linux-raspberrypi - build rtc_pcf8563 as module.

RoEdAl:
There's no reason to build this specific driver into the kernel.


kmihelich:
Why build it as a module?

RoEdAl:
`rtc_pcf8563` driver requires I2C or SPI bus driver to be loaded.
As far I know all drivers for I2C (`i2c_bcm2835`) and SPI (`spi_bcm2835`) buses are build as a module.

Despite what I wrote above PCF8563 works only on RPi 1-st gen (armv6h). On RPi 3/4 (armv7h) this driver just does not work. So feel free to ignore this PR. Sorry for bothering you.

PR #1785, opened 2020-04-03
Add community/gocryptfs

ilpianista:
`gocryptfs` does require `pandoc` only to build the man pages, so I think it's OK to add it without them.


ilpianista:
For the record, this is the output of `make build`:

```
./build.bash
gocryptfs v1.7.1; go-fuse [vendored]; 2020-04-03 go1.14.1 linux/arm
./Documentation/MANPAGE-render.bash
Rendering MANPAGE.md to gocryptfs.1
./Documentation/MANPAGE-render.bash: line 13: pandoc: command not found
make: *** [Makefile:4: build] Error 127
```

Also, commit rfjakob/gocryptfs@61940a9c should have made `pandoc` optional, but it looks like it isn't.

PR #1793, opened 2020-04-26
Add community/python-xcffib, removed haskell-xcffib dep

dequis:
The 'haskell-xcffib' package contains the code generator used to build
python-xcffib from scratch. The stable pypi releases of xcffib just
include the .py file generated by it, so they don't depend on haskell.

---

Reference: https://github.com/tych0/xcffib#installation

I chose to keep a minimal diff with the upstream pkgbuild, but honestly the python2 part stuff could be removed entirely, no one depends on that anymore. I'm also not sure why upstream isn't just using the pypi release, seems to make the whole build process more complicated for little benefit. cc @felixonmars

I guess this is what resulted in the `qtile` package disappearing from the repo. Forum thread for that: https://archlinuxarm.org/forum/viewtopic.php?f=15&t=14247

Not sure if anything special needs to be done to trigger rebuilds of `python-cairocffi` and `qtile`.

### DIFF
--- a/community/python-xcffib/PKGBUILD
+++ b/community/python-xcffib/PKGBUILD
@@ -1,0 +1,56 @@
+# Maintainer: Felix Yan <felixonmars@archlinux.org>
+# Contributor: Roger Duran <rogerduran@gmail.com>
+
+# ALARM: dequis <dx@dxzone.com.ar>
+#  - remove haskell-xcffib dependency by using the pypi release instead
+
+pkgbase=python-xcffib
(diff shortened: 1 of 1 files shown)


felixonmars:
@dequis We would like to avoid relying on generated sources if possible. Since we have a working Haskell stack it's not a big burden.

dequis:
Oh, okay, that's fair.

Thanks to you both for the reply and the merge! That was quick.

PR #1795, opened 2020-04-28
added alarm/dkms-8812au

jgmdev:
Add driver for wifi chipset RTL8812AU which was removed from 5.x kernel series and is actually being sold by companies like ameridroid and hardkernel branded as Wifi Module 5. This driver uses a github repository that has been updated to work with latest 5.x kernels, I have tested it with kernels from 4.9 to 5.7rc and it is more stable/reliable than the one shipped in older kernels.

### DIFF
--- a/alarm/dkms-8812au/dkms.conf
+++ b/alarm/dkms-8812au/dkms.conf
@@ -1,0 +1,12 @@
+PACKAGE_NAME="@PKGBASE@"
+PACKAGE_VERSION="@PKGVER@"
+BUILT_MODULE_NAME="8812au"
+BUILD_EXCLUSIVE_ARCH="armhf|armel|armv7l|arm64|aarch64"
+PROCS_NUM=$(nproc)
+[ $PROCS_NUM -gt 16 ] && PROCS_NUM=16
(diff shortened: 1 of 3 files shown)


alarm/dkms-8812au/dkms.conf
```diff
@@ -0,0 +1,11 @@
+PACKAGE_NAME="@PKGBASE@"
+PACKAGE_VERSION="@PKGVER@"
+BUILT_MODULE_NAME="8812au"
+PROCS_NUM=$(nproc)
+[ $PROCS_NUM -gt 16 ] && PROCS_NUM=16
+MAKE="'make' ARCH=arm64 -j$PROCS_NUM KVER=${kernelver}
```
kmihelich, line 6: This package is for all architectures, but the build line has ARCH=arm64?

jgmdev: Nice catch and sorry for that, I totally forgot about that work around for the wrong detection code on the Makefile, you really have super eyes! Fixed it by implementing some basic replacement logic on the PKGBUILD for aarch64 and arm64 that should always be arm64 if one of both is detected.

PR #1799, opened 2020-05-03
Add librga for Rockchip rga chip

valadaa48:


kmihelich:
What is this used with? What you have you tested it with?

This is also missing a buildarch setting to restrict to AArch64 builds.

Pick tabs or spaces in the PKGBUILD, don't use both.

valadaa48:
Hi Mike, this is a library that faces off with the CONFIG_RGA and CONFIG_RGA2 kernel drivers to provide hardware accelerated graphics operations such as scaling and rotation using Rockchip's RGA chip. We've been using this library on Ubuntu and Debian builds and I'm porting it over to ArchLinux. The package has been tested on the odroid go advance.

valadaa48:
> This is also missing a buildarch setting to restrict to AArch64 builds.

Are you referring to the arch= setting in the PKGBUILD? If so it is there however I suppose this library could be compiled for any arch (say if someone preferred to run a 32bit root instead of aarch64?)

kmihelich:
https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#arch-linux-arm-specific-packages

valadaa48:
Hi Mike, thanks I grepped the other PKGBUILDs and noticed that setting. I removed the arch= setting and didn't specify buildarch (defaults to 0) to allow for any target.

kmihelich:
Why are you removing arch? List all of the armv5, armv6, armv7, and armv8 targets you are now saying that this works with.

valadaa48:
I added aarch64 and armv7h with buildarch=12 which covers the odroid go advance and is tested for those archs. If you prefer I can amend to add all the archs.

alarm/librga/PKGBUILD
```diff
@@ -0,0 +1,42 @@
+# Maintainer: valadaa48 <valadaa48@gmx.com>
+
+_pkgname=linux-rga
+pkgname=librga
+pkgver=r36.72e7764
+pkgrel=1
+pkgdesc="Rockchip rga library"
+arch=('aarch64' 'armv7h')
+buildarch=12
```
kmihelich, line 9: Move to top.

valadaa48:
I moved buildarch to the top

valadaa48:
Sorry, I mistook your name for Mike! I fixed the comments. Let me know if you have any other questions regarding this PR. Also, sorry for the mistakes and the time it took you to review.

PR #1800, opened 2020-05-04
Add linux-odroid-go2 kernel

valadaa48:
Here is the accompanying odroid-go2 kernel. I based this PKGBUILD from linux-odroid-n2.

- Mostly defaults from def_odroidgo2
- Make dwc2 a module to allow for suspend workaround

Some other notes:

- Small gcc9 patch to mitigate 3 warnings
- Patch to rga2 that allows it to reset properly after suspend/resume.
- I'm not sure what usr/lib/modules/4.4.189-1-ARCH/build/vmlinux is used for but it adds 203MB to the package. I assume it's used for dkms?
- I removed the dep on linux-firmware as I'm not aware of any drivers needing any firmware. It's a colossal package and our devices generally have small microsd cards.


kmihelich:
Not a supported platform at this time.

valadaa48:
Thanks for your consideration. What is required to make this a supported platform? 

I run Arch on this platform as my daily driver and quite like it. Futhermore, I assume you wouldn't have any specific objections to me hosting this kernel package on my own repo?

kmihelich:
You're free to self-host.  I will need to get one of these boards at some point to bring up official support.

valadaa48:
I can see if HardKernel might be willing to send you an engineering sample if that interests you. They're releasing their rev1.1 May 11 and will be shipping engineering samples out to developers.

In the meantime I will continue to develop for this platform and in terms of core packages that will need to be upstreamed here, I don't think there are many left, if any: there are forked libdrm, ffmpeg and mpp packages which I assume provide intrinsic hw accel but so far I haven't found them to be a necessity in my workflow.

PR #1801, opened 2020-05-04
Add mali-bifrost-g31-libgl-x11-gbm

valadaa48:
Turns out the -gbm lib wasn't enough so I'm adding the x11-gbm lib as well. I re-introduced the pkgbase mechanism. I left the revision at 1 since for the original package the only thing that changed was the description (adds "(gbm)"). If you prefer that I bump the revision then I can certainly fix this.

Tested locally on my OGA and it works as expected.

Thanks!


kmihelich:
Due to build system particulars, instead of renaming please make one commit that removes the old package and then a separate commit that adds the new split package.

valadaa48:
Sure thing, will do in a min

valadaa48:
Will the build system re-build the -gbm library even though it's the same version+rel? Would be nice to ensure the pkgdesc is updated in the package.

kmihelich:
Yes, it will be rebuilt in the new split package.

valadaa48:
Perfect, thanks. Committed in 2 separate commits.

PR #1807, opened 2020-05-20
Add Odroid C4 support

jgmdev:
This pull request adds the two packages needed to support the new Odroid C4 on ArchLinuxArm which is the kernel and uboot. Since uboot requires a special old toolchain and building on x86 as with the uboot-odroid-n2 package, I uploaded a precompiled package to make things easier: 

https://github.com/jgmdev/archlinux-odroid/blob/master/uboot-odroid-c4/uboot-odroid-c4-2015.01-1-aarch64.pkg.tar.xz

I tested both packages by creating an image based on the odroid n2 and replacing the kernel and uboot. The device boots properly with ArchLinuxARM!


kmihelich:
I'm already working on these same packages, just haven't released final versions yet.

jgmdev:
I hope they can at least reduce your time investment. Just in case here is a pre-generated disk image using these packages: https://forum.odroid.com/viewtopic.php?f=203&p=293170#p293170

Cheers!

XenGi:
@kmihelich What's the staus on the official packages? Can we help you with anything?

XenGi:
@kmihelich any news? There is still no official image on the site. Is there some blocker someone could help you with?

yuvadm:
@kmihelich Would love to help out with the necessary packages for C4. It's now widely available and would be great for alarm to officially support the board.

XenGi:
@kmihelich Hello?

graysky2:
@XenGi - Be aware that the Arch ARM dev team are doing this in their spare time without financial compensation ...

yuvadm:
@graysky2 You are absolutely correct, this is why I am (and I believe @XenGi as well) offerring my assistance. Making board support a community effort is the right thing to do, in terms of spreading the work load instead of it falling in the hands of a few very skillful yet busy devs.

graysky2:
I do not doubt good intentions.  Just reminding folks who is actually making the sausage (again without a formal salary to make the sausage).  I do not know how much free-time Kevin has to spend on adding new support vs maintaining what is already there.

XenGi:
I don't demand anything from @kmihelich. But here are 2 people willing to put in the work and after he said he will do it he didn't respond anymore. I just hope he is ok and just has no time to respond.
So if he busy or smth that's totally fine, it's his free time after all. But in that case ALARM could just use the work already done by @jgmdev or let the community do the work and merge it. I don't care who commits and merges this in the end. I just want my C4 to run ALARM and if I can help making that possible I'm more than happy to do so.
At the moment I just feel blocked and I don't no why nothing happens. Does the commit just need a proper review or is their more todo? Maybe you can give us some inside in the process so we can understand better whats going on.

PR #1814, opened 2020-06-04
linux-am33x: make linux-firmware optional

ilpianista:
In Arch Linux, `linux-firmware` is optional [since 5.3.2.arch2-1](https://git.archlinux.org/svntogit/packages.git/commit/trunk?h=packages/linux&id=f1c97a49a09b0fd8d7d1c3f6e8e635ef45974373).

May we reflect the same change here?


kmihelich:
I'm guessing you've already checked that it's not using anything from there?

ilpianista:
Yes, I removed it and I'm still able to login via eth. It doesn't have WiFi nor Bluetooth, so I don't think there's anything else to check.

kmihelich:
Your comment brought up a good point.  I'll check the BeagleBones with wireless (and no ethernet) later today to make sure they're not pulling from that firmware.

ilpianista:
@kmihelich so?

kmihelich:
Sorry for the delay.  Beaglebone wireless does use firmware from this package, so not a good idea to remove it.

PR #1815, opened 2020-06-04
Introduced uboot-orangepi_r1 package and bumped uboot-sunxi to 2020.04

synytsky:
I would like to introduce Orange Pi R1 officially.
This is a tiny cheap SBC having two Ethernet interfaces.
I have two of these boards running without issues for about a year and I have successfully flashed both with the bootloader I have built with this package.
Alas, I don't have any other boards to try the rest of bootloaders that this package builds.
If you need me to update installation instructions for this board I can do that.
Basically I have used https://wiki.archlinux.org/index.php/Orange_Pi as a guide.


kmihelich:
Not a supported platform.

synytsky:
Could you please elaborate? ARMv7 is not supported?

PR #1816, opened 2020-06-18
add community/crystal for aarch64

jhass:
After this community/shards should just build and be included too, not sure what the process for that is though :)

### DIFF
--- a/community/crystal/PKGBUILD
+++ b/community/crystal/PKGBUILD
@@ -1,0 +1,98 @@
+# Maintainer: Anatol Pomozov <anatol.pomozov@gmail.com>
+# Contributor: Jonne Haß <me@jhass.eu>
+
+# ALARM: Jonne Haß <me@jhass.eu>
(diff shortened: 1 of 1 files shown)


community/crystal/PKGBUILD
```diff
@@ -0,0 +1,98 @@
...
+       CRYSTAL_PATH="$srcdir/$pkgname-$pkgver/src" \
+       CRYSTAL_CONFIG_VERSION="$pkgver" \
+       CRYSTAL_CONFIG_PATH="lib:/usr/lib/crystal" \
+       CRYSTAL_CACHE_DIR="/tmp/crystal"
+  make docs CRYSTAL_CACHE_DIR="/tmp/crystal"
+}
+
+check() {
+  cd $pkgname-$pkgver
+  return # tests require too much RAM, disable it temporary
```
jhass, line 67: Maybe not for the build environments here?

The `Makefile` also supports compiling and running the  compiler and stdlib specs separately, reducing the upper bound on used memory.

PR #1817, opened 2020-06-21
Allow uDisks2 actions to be used by Kodi

quentinus95:
When using uDisks2, the action id starts with `org.freedesktop.udisks2.`.


kmihelich:
@graysky2 review?

graysky2:
Seems consistent with what other rules written for [udisks2](https://www.freedesktop.org/wiki/Software/udisks/).

graysky2:
@quentinus95 - Thanks for the contribution, rolled into https://github.com/archlinuxarm/PKGBUILDs/commit/04f971e72d67f14a9b4344984fb6f9cd1da0a78b

PR #1822, opened 2020-07-28
Add odroid n2+ support

jgmdev:
This pull updates kernel and uboot to support odroid n2+ higher frequencies without affecting regular odroid n2 functionality. More details about the Odroid N2+ on the [hardkernel forum](https://forum.odroid.com/viewtopic.php?f=176&t=39585)


XenGi:
We could update to [odroidn2-141](https://github.com/hardkernel/u-boot/releases/tag/travis%2Fodroidn2-141) before we merge this. Thx for the work! Can't wait to get my ODROID-N2+ working. The current N2 image has a lot of issues on the N2+.

jonathonf:
This PR also switches from building from source to packaging a pre-built binary. Is that intentional?

jgmdev:
> This PR also switches from building from source to packaging a pre-built binary. Is that intentional?

Since hardkernel already has an automated build system on their github repo using travis, it uses that because the uboot branch they use requires building on x86 and one cant build it using an ARM system which complicates things as one can notice in current PKGBUILD. Since it is official uboot from hardkernel and built from travis I guess one could trust it.

jonathonf:
@jgmdev Have a look over at https://github.com/archlinuxarm/PKGBUILDs/pull/1832 and see if that works - it's a native build of the `u-boot.bin` which seems to be all that is required for the package?

jgmdev:
Great! Now all it is needed is to test if it works, it it does I think it would be a good idea to send a pull request to hardkernel so they can integrate it on their repo, maybe adding some if conditions to keep old build method and new native build method. Great job!

razaqq:
please finally merge this...

graysky2:
One reason why this hasn't been merged is that it doesn't follow the [rules](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md).

graysky2:
One potential reason why this hasn't been merged is that it doesn't follow the [rules](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md).

razaqq:
@dklementjev can you update it?

KingDuckZ:
so am I correct that we shouldn't install linux-aarch64 on the n2+ yet?

rhjdvsgsgks:
does anyone know which rule this pr didn't follow? after that we can create another pr based on this to edit it (seems like author of this pr don't want to update anymore)

PR #1823, opened 2020-08-15
alarm/uboot-rock64 to mainline 2020-07 with ATF-2.3

Kabbone:
**changes**
update uboot to mainline 2020-07
update ARM-Trusted-Firmware to 2.3
remove rockchip bootloader blobs

**resaon**
with this update rock64 also don't have any more boot problems with linux-aarch64-5.8

**additional comment**
I'm myself reduce the DRAM frequency (not in this PR) from 800MHz to 600MHz because of major stability issues with my v2 board. I don't know if we should do so in the repos or not, but it seems some v2 boards experience this issue.


rubendibattista:
This should be merged ASAP since w/ kernel 5.8.0, the old `u-boot` can't boot it. 

Ehi @Kabbone, I was reading the `u-boot` `README` about flashing into the SD card. 

```
SD Card
^^^^^^^

All rockchip platforms, except rk3128 (which doesn't use SPL) are now
supporting single boot image using binman and pad_cat.

To write an image that boots from an SD card (assumed to be /dev/sda)::

        sudo dd if=u-boot-rockchip.bin of=/dev/sda seek=64
        sync
```

You can simplify the flashing procedure just fleshing the `u-boot-rockchip.bin` that also features the SPL. Can't you?

Kabbone:
I'm not an u-boot expert, but the reason for me to splitting it up in SPL and TPL was basically that you would need that if you want to make use of the SPI flash. So if someone wants to give it a try you would already have half things in place.

If you don't care about this, you are right, it would be possible to only flash one image.

rubendibattista:
Ok, sounds reasonable. Thanks for the fast reply.

ilpianista:
Someone please remember to update http://archlinuxarm.org/os/rockchip/boot/rock64/ once this is merged (hopefully soon!)

PR #1826, opened 2020-09-03
core/linux-raspberrypi to 5.4.61-1

telsch:


telsch:
Done with this commit de6a35c

PR #1829, opened 2020-10-21
Splitting linux-firmware into a full version and stripped version

xenoxaos:
Saves ~200M of space by doing some cursory removals and then adding a linux-firmware-full package for complete firmware. Since not all kernels support compressed firmware, this is just a quick way to reduce the size of an install.


xenoxaos:
Gonna ignore this for now. Thinking of ways to only grab the firmware we can actually use

PR #1831, opened 2020-10-27
alarm/uboot-raspberrypi to 2020.10-1

chamatht:
Fixes the inability to boot from usb

### DIFF
--- a/alarm/uboot-raspberrypi/PKGBUILD
+++ b/alarm/uboot-raspberrypi/PKGBUILD
@@ -4,8 +4,8 @@
 buildarch=12
 
 pkgname=uboot-raspberrypi
-pkgver=2020.07
-pkgrel=2
+pkgver=2020.10
+pkgrel=1
 pkgdesc="U-Boot for Raspberry Pi"
 arch=('armv7h' 'aarch64')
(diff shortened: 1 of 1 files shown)


quantmint:
@kmihelich

ajc161:
I don't know if this matters but I noticed the commit for https://github.com/raspberrypi/firmware is not consistent between this and `raspberrypi-bootloader`. I updated the commit and md5sums to match here https://github.com/ajc161/PKGBUILDs/commit/251dbd055a8add5e7a9f1eb062daa31bb51bb92d. Confirmed that this still works with the Raspberry Pi 4b 8GB.

tomty89:
@quantmint Could you also add `'linux-raspberrypi4'` to `conflicts_armv7h`?

@kmihelich @graysky2 @ajc161 We should really ship dtbs (at least the ones for the aarch64 boards / that we currently ship in `uboot-raspberrypi`) in `raspberrypi-bootloader` instead. (We can either `rm` them when packaging `linux-raspberrypi{,4}`, or not make/ship any dtbs in those kernel packages at all; if we opt for the latter we simply fetch and ship all the dtbs that are available on https://github.com/raspberrypi/firmware in `raspberrypi-bootloader`.)

PR #1832, opened 2020-11-01
Build u-boot.bin for uboot-odroid-n2 using native toolchain

jonathonf:
This change builds the `u-boot.bin` for the odroid n2 using the native GCC10 toolchain. It removes the requirement for an x86 build host as well as an old non-Arch toolchain.

The resulting artefact will require some more testing to make sure it works reliably.


jonathonf:
Fails to boot.

PR #1837, opened 2020-11-29
core/linux-armv7-rc: Enable OV5640 camera driver

barni2000:
OV5640 driver is added as a module to the config.


barni2000:
I will wait for 5.11

PR #1839, opened 2020-12-15
Add community/grafana

dobo90:


community/grafana/PKGBUILD
```diff
@@ -0,0 +1,101 @@
...
+
+pkgname=grafana
+pkgver=7.3.5
+pkgrel=1
+pkgdesc='Gorgeous metric viz, dashboards & editors for Graphite, InfluxDB & OpenTSDB'
+url='https://grafana.com/'
+arch=('x86_64')
+license=('Apache')
+depends=('glibc' 'freetype2' 'fontconfig' 'gsfonts')
+makedepends=('git' 'go' 'npm' 'grunt-cli' 'python' 'nodejs-lts-erbium')
```
L0ric0, line 13: you will also need `python2` as a dep for `node-sass` and `iltorb`

L0ric0 (commented):
the `highmem` variable should be set as grafana consumes more then 2GB an build

community/grafana/PKGBUILD
dobo90: Are you sure about this? It's hard to test for me with a clean chroot on Raspberry Pi. But on x86_64 it compiles fine without python2 dependency.

dobo90:
I've pushed the second revision with highmem flag added. I've also added arm case in $CARCH which was missing.

community/grafana/PKGBUILD
L0ric0: yes as i tried to build it today on a (mostly) fresh `aarm64`. for `iltorb` it threw an error where it trys to print in python2 syntax and fails. Also it says that the error can be ignored as it is an optional dependency. For `node-sass` it says that it searches for `/usr/bin/python2` and after that searches for `python` in the path and then fails again with a syntax error on a line that trys to use python2 syntax. After installing python2 and rerunning `makepkg -csi` it worked

dobo90:
Compilation on the native armv7h machine still fails for me. I've managed to workaround some issues but in the end npm fails because it runs of out memory. I've created 4GB swap, tweaked some npm flags - still no success.
I've described everything in this commit:
https://github.com/dobo90/PKGBUILDs/commit/3b7ee22d6e4437ac3b0167dad0459b83db854184

In the other hand, I've pushed another commit into this review but it is controversial. It compiles backend side but uses frontend from precompiled x86_64 package. I've compiled it in a clean chroot, it builds fine and works on my Raspberry Pi 4.

I don't think I'll be able to do anything more with this pull request. Sorry for the hassle. If someone agrees with my approach I'll be glad. If not, please close this pull request.

L0ric0:
Is there anything that can be done to move this PR forward?

dobo90:
The only thing I could do from my side was to update this PR to latest version of grafana. I assume that it is still waiting for the review (and I'm aware of that current implementation is far away from ideal).

PR #1844, opened 2021-01-14
core/linux-kirkwood-dt to 5.10.7-1

Bonstra:
config was updated running 'make oldconfig' with all the default
answers.
Since Linux-5.10.1, arch/arm/kernel/module.lds no longer exists:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit?id=596b0474d3d9b1242eab713f84d8873f9887d980


moonman:
Thanks, linux-kirkwood-dt has been update to 5.10.8 in https://github.com/archlinuxarm/PKGBUILDs/commit/9fd38aee86009c1af123a2babc538eeef5aea65b

PR #1848, opened 2021-01-27
Bug fix for incorrectly names shared objects breaking symlinks plus g…

Feakster:
…eneral restructuring improvements

- Fixed a bug whereby the name of the shared object generated by the make command had the build target inserted.
- Added 'openmp' to the "depends" array, as it's listed as an option in the 'make' step.
- Removed 'perl' from the "makedepends" array. I presume this was added by accident for the 'sed' commands.
- Altered the "arch" array to reflect the ARM build targets.
- Moved the "Makefile.arm" erasure and build target detection to the prepare() function.
- Refactored the build target detection sequence to us if-else. The previous sequence was triggering errors where it formed the last call of a function [like prepare()]. There's probably an obvious reason for this, but I lack the bash skills to fix it, so I changed it.
- Added the 'DYNAMIC_ARCH=1' option for aarch64 build targets. This does increase the build time on aarch64, as it results in openblas being compilied for ALL ARMv8 build targets. However, having tested the use of this parameter in R, the difference in code execution time seems negligible, but this main bring performance benefits in the future.
- Restructured the package() function to more strictly adhere to library naming conventions, and to deal with the issues arising from the names of build targets being shoved in the middle of the library names for some combination of build parameters.
- Removed some unnecessary options from the 'make install' step, as they weren't doing anything.
- Removed the '-e' option from all 'sed' commands, as it's not required here (no multiple expression 'sed' calls).
- Restructured the package() function.

### DIFF
--- a/community/openblas/PKGBUILD
+++ b/community/openblas/PKGBUILD
@@ -1,5 +1,6 @@
 # Maintainer: Felix Yan <felixonmars@archlinux.org>
 # Contributor: Giuseppe Borzi <gborzi _AT_ ieee _DOT_ org>
+# Contributor: Feakster <feakster at posteo dot eu>
 
 # ALARM: Kevin Mihelich <kevin@archlinuxarm.org>
 #  - set build targets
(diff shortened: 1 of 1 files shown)


kmihelich:
I'm not maintaining an arbitrarily rewritten PKGBUILD compared to upstream.  If you have problems with how the PKGBUILD was written then take it up with the x86 Arch maintainer, not here.

Feakster:
I think 'arbitrarily' is a little unfair.

This package is currently broken on aarch64. I do not know if it is also broken on x86_64, so I don't know whether they will take any interest in my bug fixes for ARM.

PR #1850, opened 2021-01-31
core/linux-aarch64: Disable FTRACE

AlexanderStein:
Starting from 693a785f97820 FTRACE is enabled which breaks booting on
ODROID-N2. Once FTRACE is disabled again booting is back as normal.


wolfpld:
I need ftrace functionality. It would be good to periodically check if the N2 boot problem is resolved in new kernels.

AlexanderStein:
TLDR: I think this change can be reverted. It most probably is a configuration problem in the u-boot boot script.

While adding ftrace support back then did actually break booting on ODROID-N2, it was not caused by the feature itself rather than the increased size due to it. The exact same happened again later on with a different update which added new (unrelated) drivers into the kernel image. The reason is the split memory layout with a gap. My u-boot bootscript (and most probably a lot of other uses too) loaded the kernel image in the lower memory section, but as as the size increased it eventually hit the gap and wrecked havoc.

Ristovski:
Any news on this? ftrace is quite handy with all the `bcc`/`ebpf` tooling

PR #1857, opened 2021-02-23
core/linux-raspberrypi4: cmdline.txt: ttyAMA0 -> serial0

alecov:
serial0 should be used instead of ttyAMA0 in console= and kgdboc=
parameters. This fixes Bluetooth issues with newer firmware versions.

Reference:
https://github.com/raspberrypi/firmware/issues/1539#issuecomment-784498108


graysky2:
Thank you.

PR #1864, opened 2021-03-27
Add missing audio drivers for ODROID-N2 HDMI audio support

AlexanderStein:
These drivers are missing for HDMI audio support on ODROID-N2.


core/linux-aarch64/PKGBUILD
```diff
@@ -33,7 +33,7 @@ md5sums=('d2985a3f16ef1ea3405c04c406e29dcc'
          '369fb1dd626c4e0a4a7e068d27dd51c8'
          '1553fe4abbe675ed8abd4644561d8e8c'
          '9844db3484f4d4fa3e63c7e083f656f7'
-         '5e0c36c663ebe0721fb96b9f2bfef451'
+         'SKIP'
```
amstan, line 36: Please don't do this. This is the .config file, right?

amstan (changes requested):
You need to also increment pkgrel as well as per https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#arch-linux-arm-specific-packages. You might need to fetch if there were changes since you sent this pull request.

The CONFIGs are probably fine, but I'm not familair with your board. We appreciate your ample use of =m instead of =y. :)

core/linux-aarch64/PKGBUILD
AlexanderStein: Yes, sorry about that. Fix the md5sum for `.config` and adjusted the pkgrel as well.

PR #1866, opened 2021-04-15
Critical fix: extra/python to 3.9.4-1

mdevaev:
The new version of python from arch breaks the ABI and leads to segfault when using many packages, such as cython or aiohttp. Regular arch is not affected as it is a problem for 32-bit systems (in our case for armv6h and armv7h). 3.9.4 contains a hotfix for this: https://docs.python.org/release/3.9.4/whatsnew/changelog.html#changelog

The situation is made worse by the fact that cython, along with broken python, generates incorrect c code, and all packages that you rebuild from this point on will be broken and will require another rebuild. Python-aiohttp and cython was updated today and it broke too. If you don't fix this, all the cython-related packages will end up completely broken.

### DIFF
--- a/extra/python/PKGBUILD
+++ b/extra/python/PKGBUILD
@@ -12,7 +12,7 @@ shopt -s extglob
 
 pkgbase=python
 pkgname=(python python-tests)
-pkgver=3.9.3
+pkgver=3.9.4
 pkgrel=1
 _pybasever=${pkgver%.*}
 pkgdesc="Next generation of the python high-level scripting language"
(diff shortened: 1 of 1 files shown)


mdevaev:
@kmihelich sorry to ping you, but the situation seems serious enough. I wrote you an email. Due to the broken python 3.9.3, some packages that have been rebuilt since the release of this version in alarm have been corrupted due to incorrect C code. I know for sure that cython was rebuilt today and now it generates broken libraries, aiohttp rebuilt with it is also broken now. 3.9.4 will fix this problem, but will require re-assembling all binary packages (just updating python is not enough due to the broken ABI) that have been rebuilt since the release of 3.9.3.

kmihelich:
Upstream fix added in 3.9.3-1.1.  cython and dependent packages built since are rebuilding.

mdevaev:
Thanks!

PR #1872, opened 2021-05-29
core/linux-aarch64 Enable F2FS compression support

nxshock:
Can we enable F2FS compression support? Mainline Arch Linux already enabled this feature on x86-64 architecture.


nxshock:
Already done in 5b197b6168e6f9365e5066508bf264987444f1b6

PR #1873, opened 2021-06-13
Added extra/graphene to fix the crash of libgstopengl on arm / armv6h

EuphoricCatface:
Neon is "an optional instruction set extension for ARMv7 and ARMv8"
according to https://developer.android.com/ndk/guides/cpu-arm-neon.

By default, Graphene compiles with Neon instruction on ARM. A few dependency links later,
libgstopengl **in the official repo package** crashes with Illegal instruction.
Details in the commit message.

This patch probably isn't going to make 100% coverage, but it will apply to a lot of cases.

### DIFF
--- a/extra/graphene/PKGBUILD
+++ b/extra/graphene/PKGBUILD
@@ -1,0 +1,56 @@
+# Maintainer: Jan Alexander Steffens (heftig) <heftig@archlinux.org>
+
+# ALARM: Seongyong Park <EuphoricCatface@gmail.com>
+#  - Don't add NEON support where inapplicable
+ 
+pkgname=graphene
+pkgver=1.10.6
+pkgrel=1.1
+pkgdesc="A thin layer of graphic data types"
+url="https://ebassi.github.io/graphene/"
+arch=(x86_64)
+license=(MIT)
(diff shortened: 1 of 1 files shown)


EuphoricCatface:
Any updates please?

FYI, These are my attempts at debugging that led up to creating this pull request
* Archlinux ARM forum: https://archlinuxarm.org/forum/viewtopic.php?f=59&t=15269
* gst-plugins-base issue tracker: https://gitlab.freedesktop.org/gstreamer/gst-plugins-base/-/issues/909

From upstream readme says: (https://github.com/ebassi/graphene/blob/master/README.md)
> `-Darm_neon=false` - will disable the ARM NEON fast paths

moonman:
ALARM binaries only come with NEON enabled now

EuphoricCatface:
I'm just curious, if all ALARM binaries come with NEON enabled, then won't all affected binaries crash on ARM/ARMv6, since NEON is never present on those plarforms?

moonman:
I guess you missed the annauncement: https://archlinuxarm.org/forum/viewtopic.php?f=3&t=15721

PR #1874, opened 2021-06-15
 core/linux-raspberrypi4 - Enable "Intel Wireless WiFi Next Gen AGN" kernel modules

wheeskers:
This includes "Intel Wireless WiFi Next Gen AGN" kernel modules in `linux-raspberrypi4 ` package via build config change.
These modules are enabled in upstream kernel -- it would be nice to have this enabled in `raspberrypi4`-specific build as well.

Because RPi CM4 cannot boot with upstream bootloader/kernel yet so this SBC is limited to `linux-raspberrypi4` package at the moment.

Reasoning: CM4 I/O extension board has PCIe x1 slot so AX200 adapters is quite usable with it (example: https://www.jeffgeerling.com/blog/2020/wifi-6-gets-134-gbps-on-raspberry-pi-cm4).


graysky2:

wheeskers:
@graysky2
Apologies for initial formatting, PR updated. I'm not sure if this change can be considered as one that changes `pkgver` so `pkgrel` was increased by one.

graysky2:
You can just move `pkgrel` to `2`.

wheeskers:
@graysky2 
Sorry, was busy at the time of your comment. Thanks for the update!

PR #1875, opened 2021-07-22
core/linux-kirkwood-dt to 5.13.4-1

Bonstra:
Config updated with default choices from `make oldconfig` with the
exception of:
  CGROUP_MISC=y: enabled like other cgroup features; also enabled in
                 ArchLinux
  MODULE_COMPRESS_ZSTD=y: switch from gzip compression algorithm to
                          zstd (faster, similar compressed size); also
                          used in ArchLinux
  HID_FT260=m: built as module, like other HID drivers
  HID_SEMITEK=m: built as module, like other HID drivers
  CRYPTO_ECDSA=m: built as module, like the other common crypto
                  algorithms


moonman:
Please update to 5.13.5, squash commits, and I'll merge.

Bonstra:
Done!

PR #1877, opened 2021-08-26
core/linux-aarch64 to 5.13.13-1

xdarklight:
hoping that this is useful for someone


M1cha:
@kmihelich not only is 5.11 EOL, we're currently not even using the latest patch release of it. We should really update them more often.(most other linux-* packages are affected as well).

is there anything we can do to speed up the process of this? The way it is right now leads to very vulnerable systems.

xdarklight:
we now have f19887c18c3017c7af4f100aa755497d6ed2e5f1

PR #1878, opened 2021-09-03
Add extra/xf86-video-nouveau package

naortega:
These video drivers are missing from the main repos. It would seem that all that's necessary is simply to add the architecture to the `arch` list. I've only added `aarch64` since that's the only place I've tested it. To build this package it will be necessary for the server to have the key `B178BE4EA075DE07`.


davidbeauchamp:
The build system automatically overrides the `arch` array so this PR isn't necessary. `xf86-video-nouveau` is in the `armv7` already but is skipped for the other architectures as of right now.

davidbeauchamp:
`xf86-video-nouveau` has been built for `aarch64` and should be available for installation now.

PR #1880, opened 2021-09-11
core/linux-kirkwood-dt to 5.14.2-1

Bonstra:
Config updated with default choices from `make oldconfig` with the
exception of:
  BLK_CGROUP_IOPRIO=y: enabled like other cgroup features


moonman:
Thanks @Bonstra. Merged

PR #1884, opened 2021-11-13
core/linux-aarch64-rc: chromebook: do not compress dtb files

mai-000:
Samsung Chromebook Plus v1 (gru-kevin) does not boot with compressed dtb files
Resulting vmlinux.kpart is still less than 32MiB


kmihelich:
Change merged in other commits.

PR #1886, opened 2021-11-22
Update python-pypandoc to v1.6.3

marek-lach:
Update python-pypandoc from v1.4.


kmihelich:
Not supported anymore, old package for v7 has been removed now.
You also need to read the supporting documentation on rules for contributions in this repository.  This is a bad PR and would have been closed regardless.

PR #1888, opened 2021-11-24
alarm/rpi-eeprom: include needed rpi-eeprom-digest executable

jclds139:
The most recent update to `rpi-eeprom` adds another binary/script for generating signatures. This installs it alongside the rest of the package.


graysky2:
Thank you for pointing that out, updated commit and incorporated the exec you called out.

PR #1890, opened 2021-12-04
extra/mesa: enable LTO

CounterPillow:
While LTO does increase build times and memory requirements for the builder, it also results in a 20% score increase in glmark2-es2-wayland on a RK3566 Quartz64 using the panfrost driver.

I haven't seen the used memory during the build go above 2.5G so I think we're good on that front as well.


CounterPillow:
@kmihelich pretty please?

kmihelich:
Since most builders are 2GB max and LTO breaks the v7 build, this is a no for now.

CounterPillow:
Is there a way we can build only the aarch64 package with LTO? I think the improvement is significant enough that rejecting this outright is a bad idea.

PR #1891, opened 2021-12-22
Add community/yt-dlp

rien333:
Upstream builds `community/yt-dlp` with `pandoc` in makedepends, since `pandoc` is needed to build the docs. However, pandoc is hard to get by on arm(v7) machines, so we avoid it by changing the make target to only build `yt-dlp`, and nothing more, as suggested by the [yt-dlp developers](https://github.com/yt-dlp/yt-dlp#compile).

### DIFF
--- a/community/yt-dlp/PKGBUILD
+++ b/community/yt-dlp/PKGBUILD
@@ -1,0 +1,34 @@
+# Maintainer: Antonio Rojas <arojas@archlinux.org>
+# Contributor: Stefan Tatschner <stefan@rumpelsepp.org>
+# Contributor: katt <magunasu.b97@gmail.com>
+
+# ALARM: Rijnder Wever <rien334@gmail.com>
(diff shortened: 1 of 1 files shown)


CounterPillow:
You should probably squash those commits into one:

`git rebase -i HEAD~2`

Change second commit to "s", save

then `git push -f yourremote master`

rien333:
I think its just one commit now? I tend to not really do anything with git except for commiting and pushing/pulling.

CounterPillow:
Correct, you're all good now. Thanks!

graysky2:
@rien333 - Please add `zip` as a makedepends.

rien333:
Done. Let me squash again. 

(btw: I didn't add `zip` at first because, magically, it builds fine without. still, the `Makefile` (among others) does mention it in context of the yt-dlp build target, so better be safe)

PR #1892, opened 2021-12-31
added core/zstd

autumnontape:
This issue is breaking boot on non-x86_64 systems: https://bugzilla.redhat.com/show_bug.cgi?id=2035802

This PR adds a core/zstd PKGBUILD that uses a commit from a pending pull request to the zstd GitHub repository to fix it. The patched package works on my machine (passes `check()` and allows the system to boot). I assume this can't be merged exactly like this, but I'm not familiar with Arch Linux ARM protocols, so just tell me what it needs and I'll fix it up (or edit the branch yourselves if you like).


autumnontape:
@kmihelich Should this have been merged as-is? Maybe I should have been clearer about this, but I didn't update the pkgrel.

PR #1894, opened 2022-01-09
Updated core/linux-armv7 to 5.15.10

nenadperic:
Since there was no update to linux-armv7 for a while, I decided to see if I can update it myself. I have tested it on Solidrun Cubox-I2ex (i.MX6 Dual-core, 1GB RAM), no issues so far.

I'm not sure about the config changes, I tried to follow the same principle of enabling/disabling modules.

Feedback is welcome.


nenadperic:
A newer version of core/linux-armv7 is available in the repo, this one is no longer necessary.

PR #1896, opened 2022-01-19
New package: pandoc-citeproc (aarch64)

marek-lach:
Builds on aarch64 (armv8).
Crucially needed dependency to build some markdown editors.

Pandoc related packages belong to the community repository in Arch.


kmihelich:
Haskell packages aren't built here, there is no upstream package called pandoc-citeproc, and the upstream haskell-citeproc package has no dependent editor packages.

PR #1905, opened 2022-03-07
[linux-rpi] Fix for the POE+ hat fan support.

forcaeluz:
For some time my POE+ hat fan was not working (I think since the migration from linux-raspberrypi4 to linux-rpi). 
It seems that two options in the kernel configuration are not set, and that is causing the problem. After setting those to `m` all seems to work again. Tested both options individually, but only the combination of both fixed the issue for me.
Tested this on my Pi 4, with aarch64, but I assume the fix is the same for the 32bit arch.


graysky2:
Thank you for the PR.

PR #1908, opened 2022-03-15
extra/firefox: Fix regression on aarch64 (98.0.1-1.1)

marcan:
Firefox 98+ fails to draw its window on aarch64.
Upstream: https://bugzilla.mozilla.org/show_bug.cgi?id=1757571

Also add a hack to remove the psutil version cap, since otherwise it
fails to build with the latest python-psutil version (for no reason).

### DIFF
--- a/extra/firefox/PKGBUILD
+++ b/extra/firefox/PKGBUILD
@@ -11,7 +11,7 @@ highmem=1
 
 pkgname=firefox
 pkgver=98.0.1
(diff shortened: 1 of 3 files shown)


kmihelich:
Thanks for tracking down the issue!

rmueller83:
Looks like this will be the proper patch: https://github.com/crossbeam-rs/crossbeam/pull/802

PR #1913, opened 2022-03-27
extra/mesa: add rockchip_ebc support

Strit:
Mesa already has a driver for the EBC interface found
in some rockchip devices, like the PineNote.
We just need to tell mesa to build it.
So add a patch from @0cc4m that does this.
Hopefully @0cc4m will upstream it to upstream Mesa.

Signed-off-by: Dan Johansen <strit@manjaro.org>


Strit:
Just for information.

The linux kernel does not have proper support for EBC yet, but it is being worked on. So this is just to have the mesa part ready for it.

PR #1914, opened 2022-03-29
Compile jemalloc with 16 KiB page size on arm

Diatrus:
This is needed for kernels with 16 KiB page sizes (currently the kernel for the Apple M1).
This was tested on a 16 KiB kernel does not cause any functionality regression.

See the following discussion for why page size must be set higher or equal to expected:
https://github.com/jemalloc/jemalloc/issues/467#issuecomment-294383178

### DIFF
--- a/extra/jemalloc/PKGBUILD
+++ b/extra/jemalloc/PKGBUILD
@@ -1,0 +1,46 @@
+# Maintainer:  Bartłomiej Piotrowski <bpiotrowski@archlinux.org>
+# Contributor: Massimiliano Torromeo <massimiliano.torromeo@gmail.com>
+# Contributor: Kovivchak Evgen <oneonfire@gmail.com>
+
+# ALARM: Hayden Seay <me@diatr.us>
+#  - build with 64 KiB page sizes on arm.
+
+pkgname=jemalloc
+epoch=1
+pkgver=5.2.1
+pkgrel=6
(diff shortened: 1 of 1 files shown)


kmihelich:
How about on a 4k kernel? Everything here is currently 4k.

Diatrus:
> How about on a 4k kernel? Everything here is currently 4k.

According to the linked comment and subsequent reading of the actual configure argument, it should work on 4K. However, I do not personally have a device that currently supports a 4 KiB page size kernel and would appreciate help with a proper test in that regard.

jannau:
Using jemalloc compiled for a larger page size than the system's page size is not problem. rust with static copy of jemalloc work with this [patch](https://github.com/AsahiLinux/alarm-PKGBUILDs/commit/c2459a0ae6fc04b7fe98bb04f10795248eca949b) changing the page size to 64k on systems with 4k and 16k kernel page size without problems.

leifliddy:
So it looks like Fedora sorted out this issue the same way four years ago by including the following logic in their version of a PKGBUILD
```
# Override PAGESIZE, bz #1545539
%ifarch %ix86 %arm x86_64 s390x
%define lg_page --with-lg-page=12
%endif

%ifarch ppc64 ppc64le aarch64
%define lg_page --with-lg-page=16
%endif
```

https://bugzilla.redhat.com/show_bug.cgi?id=1545539

PR #1915, opened 2022-03-31
aarch64: enable BTF and disable unprivileged BPF

fntlnz:
- BTF was enabled to allow CO-RE programs to run on aarch64 machines.
- Unprivileged BPF was disabled by default to reduce the attack surface
  of the bpf() syscall

If there are any questions I'm happy to answer. I can port the changes to the other boards as well if there is interest.


kmihelich:
Only linux-aarch64 is necessary at this time, the next rc kernel will be seeded from that.

Why is DEBUG_INFO enabled?

fntlnz:
@kmihelich ah that makes sense, I'm removing the -rc one.

DEBUG_INFO is a prerequisite for `CONFIG_DEBUG_INFO_BTF` to work, this should add about 2 megabytes to the final vmlinux file.

Notice that I'm not enabling DWARF, which could add 100s of megabytes to the final kernel.

For reference, Arch Linux ships with the same config: https://bugs.archlinux.org/task/66260

fntlnz:
it looks like there's no interest in making this happen, I'll close this

PR #1918, opened 2022-04-17
core/linux-aarch64: enable Quartz64 A support

CounterPillow:
This enables a few config options to support the Quartz64 Model A board by PINE64.

Notably, `CONFIG_MOTORCOMM_PHY` needs to be y instead of m to stop the generic PHY driver from being loaded before the motorcomm one. Alternatively, if it is built as a module, it needs to be included in the initramfs.

PCI-e host support is enabled though the PHY for it is in 5.18 so this won't work yet, but will be needed a release from now.


CounterPillow:
Was apparently quietly added to the configs during the 5.18 bump so I can close this now. Thanks!

PR #1922, opened 2022-05-24
extra/libunwind to 1.6.2-2

jannau:
Adds the merged but not yet released upstream PR #330
(https://github.com/libunwind/libunwind/pull/330). Fixes Xorg on Apple
silicon with 16k page size.


marcan:
Note that this regressed in a recent update and is breaking Xorg for everyone, so this is somewhat urgent :)

PR #1923, opened 2022-06-06
Add community/darktable

mipimipi:
darktable is not yet available for Arch Linux ARM.

The package builds with unchanged PKGBUILD on aarch64 (tested in a clean chroot). It doesn not build on armv7h.

The only changes to PKGBUILD is setting the ALARM-specific parameters `buildarch` and `highmem`.


mipimipi:
By accident I created a dependency between #1923 and #1924. I will create new PR's.

PR #1924, opened 2022-06-07
Fix for Apple M1 in extra/rust

mipimipi:
On Apple M1, the rust package does not build.

This PR sets the page size for jemalloc to 16K (as originally [suggested](https://github.com/AsahiLinux/alarm-PKGBUILDs/commit/c2459a0ae6fc04b7fe98bb04f10795248eca949b) by @marcan). With that fix, the rust package can be built and used on Apple M1. Apart from Apple M1, I tested the fix for "standard" aarch64 and for armv7h successfully.


mipimipi:
By accident I created a dependency between #1923 and #1924. I will create new PR's.

PR #1926, opened 2022-06-09
Add community/darktable

mipimipi:
darktable is not yet available for Arch Linux ARM.

The package builds with unchanged PKGBUILD on aarch64. It does not build on armv7h. The only changes to PKGBUILD is setting the ALARM-specific parameters `buildarch` and `highmem`.


Strit:
Usually when a package builds directly from ABS sources, there's no need for a PKGBUILD here.

Maybe the reason why it's not in Arch Linux ARM is that it does not build on armv7h as you mention.

Have you looked into why it doesn't build?

PR #1928, opened 2022-07-21
Add support for 16k pages in qtwebengine

yuyichao:
This is the backport of https://github.com/chromium/chromium/commit/a704c3a94179e6f3abb28f090d096ced72657d7c and is needed to make qtwebengine works on apple m1 hardware.


yuyichao:
@kmihelich Thanks.

Note that in qt6.4 the webview part of the patch can be dropped since it'll presumably be chromium 102 based (https://wiki.qt.io/QtWebEngine/ChromiumVersions has not been updated yet). However, the pdfium part of the patch still needs to be applied. Ref https://bugreports.qt.io/browse/QTBUG-105145

PR #1929, opened 2022-07-25
add ARMv8 deprecated config options

dkadioglu:
...to allow, that 32bit armhf binaries can run again with this 64bit kernel. Arch Linux ARMs linux-aarch64 has them, too.
Without those kernel options, I cannot use the Raspberrymatic Docker container, which runs fine on armv7 but not on aarch64, as some 32bit binaries have to be run in this container and they fail to run. According to https://github.com/jens-maus/RaspberryMatic/commit/c42bcb218fb8d1a6f44a9c6a6c9a6085ebdec3bb these are needed to allow running 32bit armhf binaries on arm64. I successfully built the kernel with those options and successfully runtime tested it on my RPi 4B 8GB. The Raspberrymatic Docker container runs fine on aarch64 linux-rpi now, also the rest of my system runs fine as before so far.
If there are any questions or further info needed, I am happy to follow up. I would be thankful, if this small modification could be considered, when there is nothing against it.


graysky2:
Thank you for the PR, merged in https://github.com/archlinuxarm/PKGBUILDs/commit/ded981b1f02ac60a5aacb313e5429b2869829d6e

PR #1932, opened 2022-08-08
added community/earlyoom

maximbaz:
Hello, I'm going through the packages I maintain to make sure as many of them can also be built for ARM.

This one optionally depends on `pandoc`, thus the only change.

I also made some changes directly in Arch [community] to `wldash` and `kubernetes` (`kube-apiserver` `kube-controller-manager` `kube-proxy` `kube-scheduler` `kubeadm` `kubectl` `kubelet` `kubernetes-control-plane-common`) that let them compile successfully on ARM (those changes are trivial, don't hurt to have in main repos, and release some maintenance burden from you), would those be automatically picked up and compiled for ALARM?

### DIFF
--- a/community/earlyoom/PKGBUILD
+++ b/community/earlyoom/PKGBUILD
@@ -1,0 +1,36 @@
+# Maintainer: Massimiliano Torromeo <massimiliano.torromeo@gmail.com>
+# Maintainer: Maxim Baz <$pkgname at maximbaz dot com>
+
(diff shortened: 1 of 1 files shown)


kmihelich:
Thanks for making the changes upstream, that does help me out a lot.  Upstream changes not tracked here will get automatically picked up and built.  wldash built successfully but kubernetes failed, I'll have to look into that.

PR #1942, opened 2022-09-14
extra/thunderbird: Fix compilation and update to 102.2.2

Strit:
Newer versions of Thunderbird has trouble compiling with system python packages. So switch to using pip.

Signed-off-by: Dan Johansen <strit@manjaro.org>


kmihelich:
Packages here never exceed upstream Arch versions.

PR #1949, opened 2022-10-18
core/linux-* fix patching silently failing when in-tree

rmsc:
Patches were being applied with `git apply` despite the kernel tarball not being a valid git repo. When building in-tree, git would get confused with the repository root directory, and patching wouldn't be done. The resulting package would therefore be incorrect.

This replaces all calls to `git apply` with calls to `patch`.


rmsc:
I didn't update the pkgrel because I assumed that there's no change for CI-built packages. If that's not the case please let me know and I'll add that change.

moonman:
https://github.com/archlinuxarm/PKGBUILDs/pull/1927

PR #1950, opened 2022-10-19
Add support for the Odroid-M1

rmsc:
This adds support for the Odroid-M1 SBC from Hardkernel, by applying the following upstream patchset:
https://patchew.org/linux/20220930051246.391614-1-aurelien@aurel32.net/

All the required support and device drivers for the SoC are already in 6.1-rc1, but unfortunately the patchset adding the device tree files for this particular SBC missed the merge window.

Not sure if merge rules will allow it to be included in rc2 or not, but for now this is working fine without any special kernel config change.

This obviously depends on #1948.


rmsc:
I've also enabled a few drivers that were disabled by default. The VOP2 is required for HDMI output.

refutationalist:
I have compiled this kernel on a Honeycomb LX2 and can confirm it does properly boot on an ODROID-M1.

bswartz:
Now that mainline Linux boots on the M1, is this needed anymore?

PR #1953, opened 2022-10-27
extra/qemu: make qemu-base to work on aarch64

adam900710:
Since I'm one of the very few users using ArchlinuxARM and running KVM (libvirt) on it, it's really a pity that ArchlinuxARM is unable to easily install qemu.

There are several problems involved in the current situation:

- Bad qemu-base/desktop dependency
  They still depends on x86 target, which is completely wrong on aarch64 systems.

- Hard requirement on firmwares
  This is especially bad for Aarch64, as we can not build edk2-virtarm natively right now.
  (But we can easily install the same package from upstream archlinux edk2-virtarm, as they are just firmwares)

- Bad libbpf dependency
  This looks like the package is not properly linked to the current libbpf

- Link failure for qemu-static
  This happens with latest native toolchain.

Currently I modified the PKGBUILD to handle above problems by:

- Make qemu-base/desktop to depend on qemu-system-aarch64
- Make edk2-* and seabios dependency optional
- Make the PKGBUILD work at least
- Disable user-static builds completely

So far it at least compiles and qemu can at least start up.

### DIFF
--- a/extra/qemu/PKGBUILD
+++ b/extra/qemu/PKGBUILD
@@ -23,15 +23,15 @@ pkgname=(
   qemu-tests
   qemu-tools
   qemu-ui-{curses,dbus,egl-headless,gtk,opengl,sdl,spice-{app,core}}
-  qemu-user{,-static}{,-binfmt}
+  #qemu-user{,-static}{,-binfmt}
   qemu-vhost-user-gpu
   qemu-virtiofsd
   qemu-{base,desktop,emulators-full,full}
 )
 pkgver=7.1.0
-pkgrel=10
+pkgrel=11
 pkgdesc="A generic and open source machine emulator and virtualizer"
-arch=(x86_64)
+arch=(x86_64 aarch64)
 url="https://www.qemu.org/"
 license=(GPL2 LGPL2.1)
 # TODO: consider providing rdma-core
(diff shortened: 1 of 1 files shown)


extra/qemu/PKGBUILD
```diff
@@ -23,15 +23,15 @@ pkgname=(
...
   qemu-vhost-user-gpu
   qemu-virtiofsd
   qemu-{base,desktop,emulators-full,full}
 )
 pkgver=7.1.0
-pkgrel=10
+pkgrel=11
 pkgdesc="A generic and open source machine emulator and virtualizer"
-arch=(x86_64)
+arch=(x86_64 aarch64)
```
clarfonthey, line 34: 😬

adam900710: What's the problem?

I'm building the whole qemu group natively thus adding the arch.

clarfonthey: Sorry, this is more me being uncomfortable that the archlinuxarm PKGBUILD for qemu was copied over without even changing the arch, which makes me wonder what other packages are configured incorrectly like this. Does the builder just ignore the arch?

adam900710: Oh, I see. The Alarm modifies its build systems to mostly ignore the `arch` specification.

I just hate the extra warning when running "makepkg".

You can ignore the change completely and accept the warning.

adam900710:
Closed as I have a better version to automatically switch dependency for qemu-base and qemu-desktop.

PR #1954, opened 2022-11-02
Fix boot for RaspberryPi 4

dvzrv:
alarm/uboot-raspberrypi/boot.txt.v3:
Using `booti` with RaspberryPi 4 is broken if providing the address via `fdt_addr_r` and not `fdt_addr`.


kmihelich:
No issues on my RPi 4.  Your change is not a fix.  Line 9 is loading the kernel's dtb into fdt_addr_r.  I have no idea what dtb you're using in this instance.

dvzrv:
My apologies, this was supposed to target RPi400, not 4.

There I see:

```
mmc1: unrecognized SCR structure version 2
mmc1: error -22 whilst initializing SD card
mmc1: unrecognized SCR structure version 4
mmc1: error -22 whilst initializing SD card
```

I guess this means not using a dtb fixes the boot? :thinking:

dvzrv:
Even providing the specific file (i.e. `/dtbs/broadcom/bcm2711-rpi-400.dtb`) to the load command (instead of relying on `${fdtfile}`) leads to the same result.

Not providing a variable to [booti](https://u-boot.readthedocs.io/en/latest/usage/cmd/booti.html) also seems to be no option, as it requires the `fdt` parameter.

dvzrv:
Looking at the [u-boot documentation on environment variables and image locations](https://u-boot.readthedocs.io/en/latest/usage/environment.html#image-locations) it is a bit unclear what is used when and why... and why this approach even works... oO

kmihelich:
I assume the fdt_addr is being populated with some stock dtb provided in the firmware prior to the U-Boot load of the upstream kernel's dtb load.  But I also don't have a 400 to test and develop against locally.  Since some otherwise provided dtb is working and the linux-aarch64 package's dtb isn't, I'm at a bit of a loss at the moment.  The kernel itself is clearly fine since it's booting with one dtb, but I don't know why it's failing with the mainline kernel dtb.

Mathspy:
I can confirm that this fix works for my Pi 4. See this thread: https://archlinuxarm.org/forum/viewtopic.php?f=67&t=15422&start=20#p67299
It seems the issue has nothing to do with whether the Pi is 4 or 400 or even CM4. There was a recent change in hardware that breaks the current installation
Was affecting my Pi 4 (which I just bought this weekend) and following the fix in the forum thread above made it finally work for me.
I don’t really know if this is the right fix or not but it’s definitely a decent temporary fix, I wonder if it would be possible to apply it conditionally depending on the revision of the Pi.

mhabb:
+1, had the same issue on my Pi 4 and this change fixed it.

lategoodbye:
As already figured out this is a u-boot issue:
https://lists.denx.de/pipermail/u-boot/2023-January/504038.html

goekce:
I can also confirm that without this patch:

- the B0 revision of the ARM chip on RPi 4 works fine on aarch64
- but C0 *does not*.

I tried to follow the thread but do not understand if the issue is fixed or not:

https://lists.denx.de/pipermail/u-boot/2023-January/504062.html

lategoodbye:
> I tried to follow the thread but do not understand if the issue is fixed or not:
> 
> https://lists.denx.de/pipermail/u-boot/2023-January/504062.html

Hi @goekce, are you sure about this link? NAND has nothing todo with this issue. Didn't followed any recent u-boot discussions about this topic.

goekce:
If you follow the thread that you linked using the *next message in thread* link, then you arrive there.
You are right. The message that I linked has nothing to do with this issue, I just wanted to be sure.

PR #1957, opened 2022-11-17
Add alarm/raspberrypi-overlays for UEFI firmware

arenekosreal:
This is required by uefi-raspberrypi4, it is the /boot/overlays folder of the Raspberry Pi 4.

This should be provided by linux-raspberrypi, but UEFI firmware needs linux-aarch64 for the devicetree, besides, linux-raspberrypi does not enable ACPI and UEFI related config so it is not suitable for uefi-raspberrypi4.


arenekosreal:
Closing as not needed by `uefi-raspberrypi4` anymore.

PR #1959, opened 2022-11-24
extra/qemu: to make qemu installable for aarch64

adam900710:
Currently qemu package group has its problems, making it almost impossible to run libvirt with qemu:

- qemu-user-static build will fail using native tool chain This will prevent the whole package, including qemu-system-* which can compile without any problem.

- qemu-base and qemu-desktop depends on qemu-system-x86 Which makes completely no sense. This is something from the Archlinux PKGBUILD.

- No way to install qemu-system-aarch64 As it has hard dependency on edk2-armvirt, which is not buildable yet natively on aarch64 (missing arm toolchain AFAIK).

This update will solve the problems by:

- Split qemu-user-static into its own PKGBUILD Qemu-user-statis has a completely different use-case compared to system emulators. The most common usage is allowing the host to chroot into rootfs for another arch.

  Thus it is not suitable to contain qemu-user-static into qemu package. There is already a feature request to upstream Archlinux: https://bugs.archlinux.org/task/76556

- Make qemu-base and qemu-desktop to change its default system emulator dependecy Now the PKGBUILD will choose the default system emulator depending on the target arch.

- Make edk2-armvirt as optional dependency In fact we can directly install the same package from upstream Archlinux, as that package is just the edk2 uefi firmware, which is arch independent.

Have tested qemu PKGBUILD natively on both x86_64 and aarch64, and resulted qemu can work with libvirt using KVM.

### DIFF
--- a/extra/qemu-user-static/PKGBUILD
+++ b/extra/qemu-user-static/PKGBUILD
@@ -1,0 +1,138 @@
+# Maintainer: David Runge <dvzrv@archlinux.org>
+# Contributor: Tobias Powalowski <tpowa@archlinux.org>
+# Contributor: Sébastien "Seblu" Luttringer <seblu@seblu.net>
+# Contributor: Qu Wenruo <wqu@suse.com>
+
+pkgbase=qemu
+pkgname=(
+	qemu-user-static{,-binfmt}
+)
+pkgver=7.1.0
+pkgrel=99
+pkgdesc="QEMU static user mode emulation"
+arch=(x86_64 aarch64)
+url="https://www.qemu.org/"
+license=(GPL2 LGPL2.1)
+makedepends=(
+	meson
+)
+source=(
+  https://download.qemu.org/qemu-$pkgver.tar.xz{,.sig}
+  $pkgbase-7.1.0-io_uring.patch::https://patchwork.kernel.org/project/qemu-devel/patch/20220924144815.5591-1-faithilikerun@gmail.com/raw/
+  $pkgbase-7.1.0-block1.patch::https://gitlab.com/qemu-project/qemu/-/commit/a7c5f67a78569f8c275ea4ea9962e9c79b9d03cb.diff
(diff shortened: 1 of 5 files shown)


Rippanda12:
does not compile qemu-user-static
```/usr/bin/ld: (.text+0x10): warning: too many GOT entries for -fpic, please recompile with -fPIC```

adam900710:
@Rippanda12 Static build failure is the reason to split that package out.
No matter if using the old or the new PKLGBUILD. What I do here is to make the qemu-system-* work as the primary objective.

And I have already mentioned, `Have tested qemu PKGBUILD natively on both x86_64 and aarch64`, that only means the PKGBUILD for qemu, not qemu-user-static.

I only care about the qemu-system-* for now.

Feel free if you can fix the static part.

adam900710:
Furthermore if your objective is mostly to run programs of another arch, like chroot into a rootfs of another arch, then qemu-user-* with qemu-user-binfmt is better.
And with the new PKG, qemu-user can already compile without problem (inside the old qemu group).

Qemu-user-static used to be the goto solution, but that's before the binfmt package.
Before binfmt, one may need to copy the static build into the target rootfs, and chroot with `chroot /usr/lib/qemu-static-aarch64 /bin/bash`.

Another thing is, there isn't much usage for qemu-user* (no matter static or nor) AFAIK.
Under most cases, aarch64 is the weaker target, thus we want qemu-user* on a stronger host (mostly x86_64).

The only exception is Apple M1, which may want qemu-user-* to access/run/chroot into x86_64 world.

But anyway, with the new non-static only PKGBUILD, you don't really need qemu-user-static anymore.
Or did I miss something?

adam900710:
For the static build failure, it's in fact the problem of the glibc package.
https://gitlab.com/qemu-project/qemu/-/issues/1340#note_1186317121

In fact the newer glibc should be compiled with `-fPIC` now, which means ArchlinuxARM should also update its glibc first.

Strit:
Seems some of these changes where added recently:

https://github.com/archlinuxarm/PKGBUILDs/commit/91c6b8c471c8b89216d8163f6544d475216f019c

PR #1968, opened 2022-12-21
core/linux-aarch64: enable config options for RK356x

CounterPillow:
The RK3566 and RK3568 already boot with this kernel, but a few things are missing due to left out config options. The biggest one will be video output (VOP2), followed by the PCIe/USB/SATA combination PHY (NANENG). While we're at it, also enable the CSI PHY for this SoC and other SoCs using the same PHY.


felixsinger (approved):

felixsinger:
@kmihelich Could you take a look please?

cbix:
@CounterPillow this is all upstream already, you can close the PR :)

PR #1969, opened 2022-12-22
alarm/devtools-alarm: Remove -d option from pacstrap

CounterPillow:
An upstream update dropped the deprecated -d option from pacstrap, which broke mkarchroot in devtools-alarm. Apply the upstream devtools patch to fix this issue.

Bug report reference: https://archlinuxarm.org/forum/viewtopic.php?f=15&t=16282

### DIFF
--- a/alarm/devtools-alarm/0010-remove-deprecated-d-option.patch
+++ b/alarm/devtools-alarm/0010-remove-deprecated-d-option.patch
@@ -1,0 +1,30 @@
+From a510331698bb13ca11ccd7db3f9beef75f87744c Mon Sep 17 00:00:00 2001
+From: Morten Linderud <foxboron@archlinux.org>
+Date: Mon, 21 Nov 2022 10:51:02 +0100
+Subject: [PATCH] mkarchroot: remove deprecated -d option
+
+It's been deprecated for a few years now.
+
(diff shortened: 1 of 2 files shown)


CounterPillow:
@kmihelich I know you are very busy with keeping on top of the latest kernel updates, but could you take a look at this?

CounterPillow:
@graysky2 what is wrong with this PR? :slightly_smiling_face:

CounterPillow:
I know, but you frequently claim that PRs are left open indefinitely because there's something wrong with them
image
image

Since you appear to be an expert on this topic, I thought you could either shine some light on what's wrong or stop making claims like this that dismiss my contributions.

graysky2:
A common reason for PRs being ignored is not following the rules.  @kmihelich is the maintainer of packages for which you opened PRs.  I cannot speak for him.

SpieringsAE:
well this has now been open for 2 years, and it breaks a very regular maintainer workflow. It would be very nice to get this in. What would be even better is to update devtools in general since it is still based on a very old arch work flow.

One of the big things holding this up in my view is the ghc/ghc-static packages, these seem to have a dependency cycle that I have not yet found an entrance into. At archriscv they seem to have it figured out but I still cannot find the entry point in their package builds.

PR #1971, opened 2022-12-28
extra/mesa: Re-enable LTO for aarch64

CounterPillow:
LTO was disabled in 06c1615 without an explanation, so I assume it wasn't done intentionally, as the comment mentioning LTO was left in the PKGBUILD.

Bug report reference:
https://archlinuxarm.org/forum/viewtopic.php?f=15&t=16285


Strit:
I believe LTO was disabled upstream by https://github.com/archlinux/svntogit-packages/commit/45fa5bb1460b64fb3fe5ff36451bea8976bf975a and has not been enabled upstream since then because of a [rendering issue](https://bugs.archlinux.org/task/76019?project=1&string=mesa).

However, I have built Mesa 22.3.2 with LTO enabled on aarch64 and tested it on my Pinebook Pro. No rendering issues.
So the issue might be [AMD specific](https://gitlab.freedesktop.org/mesa/mesa/-/issues/7352).

Should be safe to re-enable LTO build on aarch64 again.

PR #1972, opened 2023-01-04
core/linux-armv7 to 6.1.2

MastaG:
This updates the linux-armv7 kernel to stable 6.1.2:

- Add and rebase RCN 6.0.rc5 patch
- Add ODroid XU3/XU4 patches for Panfrost and MFC video decoding
- Add Hardkernel edid firmware binaries
- Rebase kernel config

### DIFF
--- a/extra/valgrind/PKGBUILD
+++ b/extra/valgrind/PKGBUILD
@@ -21,7 +21,7 @@ noautobuild=1
 
 pkgname=valgrind
 pkgver=3.19.0
-pkgrel=5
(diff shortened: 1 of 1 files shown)


MastaG:
Can somebody please review this?
Because with the patchset we can run the mesa panfrost driver (at full speed) on ODroid XU3 and XU4.
All packages including the one for chromebook at odroid c1 are built successfully.

Screenshot from 2023-01-06 16-13-43

PR #1974, opened 2023-01-08
Include chromedriver in chromium package

cryzed:
See [here](https://archlinuxarm.org/forum/viewtopic.php?f=9&t=16306) for more information. I'm not sure why this was removed initially, but I successfully compiled the chromedriver binary on my Arch Linux ARM (Raspberry Pi 4) system and it works without issues. It's even compiled by default in the PKGBUILD, just not packaged for some reason.


cryzed:
Can I do something to make this easier to merge for you guys? Did I do something wrong?

PR #1977, opened 2023-01-14
graphene 1.10.8-1: disable neon simd on armv7h

MastaG:
This disables neon simd on armv7h devices only, because it crashes otherwise.
See: https://github.com/ebassi/graphene/issues/215
Redhat does the same thing in their spec file.

This is required to get the gnome-shell desktop running again om armv7h devices.
One would need to build and push an updated gnome-shell package though (it builds fine here).

### DIFF
--- a/extra/graphene/PKGBUILD
+++ b/extra/graphene/PKGBUILD
@@ -1,0 +1,60 @@
+# Maintainer: Jan Alexander Steffens (heftig) <heftig@archlinux.org>
+ 
+pkgname=graphene
+pkgver=1.10.8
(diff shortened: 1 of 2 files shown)


MastaG:
@kmihelich 
Could you please consider merging this?
Almost all ARMv7 distributions are disabling neon in graphene as well.
See the discussion here: https://github.com/Igalia/meta-webkit/pull/448

kmihelich:

MastaG:
Right, I'll take care of it next week

PR #1978, opened 2023-01-14
xdg-desktop-portal 1.16.0-1: revert back to autoconf

MastaG:
Build xdg-desktop-portal with autoconf instead of meson.
This is required because meson doesn't detect correct string size for armv7h causing segfaults when loading the libraries.
No other changes have been made.


a-andre:
Could you provide a link to the upstream bug report of the issue you try to work around.

BTW, you have to update `makedepends`.

MastaG:
Good catch @a-andre thanks :)
I haven't reported the issue to the upstream project yet.
Will do this in the following days and update this PR.
I could however reproduce it locally, where building with meson seems to produce broken binaries which crash instantly.

MastaG:
https://github.com/flatpak/xdg-desktop-portal/issues/952

PR #1979, opened 2023-01-14
Enabling TPM drivers in linux-aarch64

mys721tx:
Enables TPM drivers in linux-aarch64 to allow the use of VMWare vTPM.


graysky2:
This has already been implemented, closing as redundant.

PR #1983, opened 2023-01-25
core/linux-aarch64: Update to 6.1.8 (current stable)

lynix:
The current linux-aarch64 package version ships a kernel (v5.19.8) that is more than 4 months old.

The 5.19 branch is EOL as it's not an LTS version and there seem to be [multiple CVEs](https://www.linuxkernelcves.com/streams/5.19). This pull request updates to the current stable release (v6.1.8).

Built and tested on an Odroid C4.

@kmihelich: If you don't like b21333e then please feel free to drop that one and only take the version bump, or tell me to send a pull request without it.


CounterPillow:
Judging by my experience submitting PRs to this repo, I doubt you'll get a response to this, but I hope I'm wrong and we finally get a new kernel.

lynix:
To be honest I've got not much hope either. But you can only complain if you have tried, so here we go. I've had a pull request accepted in the past so it's not completely impossible.

I'd really appreciate having such a critical component be maintained again. Or is `linux-aarch64` regarded as an unofficial kernel package for unsupported devices? In that case I'd wish that it was written somewhere.

The description right on the distro front page unfortunately reads like a bad joke (for this package):

> New software versions are packaged as they are released, ensuring you are always on the leading edge of stable software releases.

mrpllewellyn:
Seconded, if it is an unofficial kernel for unsupported devices it would be useful to know this

lynix:
As v6.1.9 has been released, I will close this PR and open a new one, starting with that version.

This time I will set it up so that it can be updated for newer versions so that I won't need to open a new one for each new kernel release.

PR #1985, opened 2023-02-03
Enable Arm-specific features of pixman

akihikodaki:
The lack of Arm-specific features do not only result in bad performance, but causes behavioral changes as there is no fallback for pixman_blt(), for example.


kmihelich:
Fixed in 63b3906

PR #1987, opened 2023-02-26
added community/dtools

CyberShadow:
[Upstream](https://github.com/archlinux/svntogit-community/blob/packages/dtools/trunk/PKGBUILD) uses `dmd` directly, which can only target `x86` and `x86_64`.

This version uses `ldc` which can target more architectures, including ARM.


amstan:
My bad @CyberShadow, I mislead you. Please keep it in one commit, but say what [you changed in a header in the PKGBUILD](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#upstream-x86-arch-linux-packages) (exercise to find similar comments in other PKGBUILDs left to the reader).

PR #1992, opened 2023-04-20
Enable Rados block device kernel module on linux-rpi packages.

Chief-Detektor:
Compiles fine and module loads and works as expected.
Since it's compiled in the linux-aarch64 packages as well I think makes sense to include it in the linux-rpi packages too.


Chief-Detektor:
I tested it on an rpi-4b and an rpi-3b (64 Bit)
I miss testing it on an arm7hf. I will give feedback again once I've tested that.

Chief-Detektor:
Compiled and loaded the rbd kernel module on an rpi-2b (armv7l).
Works!

graysky2:
Implemented in https://github.com/archlinuxarm/PKGBUILDs/commit/6e525b557f43d4f1e7a6d2c820f93fa7f390036a

PR #1995, opened 2023-05-17
extra/mesa to 23.1.0-0

akaWolf:


davidbeauchamp:
We don't go above upstream's current version, when mesa is released there it will get bumped here as well.

1ace:
@davidbeauchamp mesa 23.1.0 was released a week before this PR, I just forgot to update the website which is probably what you checked; sorry about that :upside_down_face:

One note for @akaWolf though: asahi is **not** usable upstream, please don't enable it as it will only fail to load for users (by design) :wink:

davidbeauchamp:
@1ace upstream, in this case, is Arch Linux. We track their package version https://archlinux.org/packages/extra/x86_64/mesa/, which is still at `23.0.3-1`. When mesa is bumped in mainline Arch, it will get bumped in Arch Linux ARM.

PR #1998, opened 2023-06-05
extra/bitcoin: drop Protobuf & update Changelog

fanquake:


fanquake:
I see this is actually just tracking upstream. Will push the changes there instead.

PR #2000, opened 2023-06-28
core/linux-armv7 enable dtb symbols

dacioromero:
Similar to #1623, linux-armv7 needs symbols on the /boot/dtbs for overlays to work. Tested on my Banana Pi M2 Zero and the sun8i-h3-uart3.dtbo from ripped from Armbian and m2-zero-eth0.dtbo from [this blog post](https://lisy.dev/banana-pi-m2-zero.html).


dacioromero:
Thank you @graysky2!

PR #2002, opened 2023-07-17
Deno

wirewc:
**Addition:** Adds necessary changes for building natively on ARM targets for [deno](https://deno.land). 

Tests:
- [x] Build tested
- [x] Builds in clean chroot
- [x] Tested on a live M1 Mac Mini running Asahi Linux and builds on an Odroid N2.


wirewc:
@kmihelich This is needed as a dependency for another package coming for ALARM.

wirewc:
Pining @amstan per request on IRC. Thank you for helping earlier!

PR #2004, opened 2023-08-10
core/systemd to 254.1

7Ji:
Port of https://gitlab.archlinux.org/archlinux/packaging/packages/systemd/-/commit/a5ee36bfd27adf58c419b596cbbc27c293640e97

This update is urgent as it contains an fix to https://github.com/systemd/systemd/issues/28671

Version 254 contains a bad /usr/lib/udev/rules.d/60-persistent-storage.rules file which will cause `ID_NAME` and `ID_SERIAL` fields missing for MMC/Memstick block devices and `/dev/disk/by-id/mmc*` links to not be created. This will break a user's boot configuration or fstab if they're using `/dev/disk/by-id/ mmc*` links as rootfs identifier or fstab mounting sources.

As most alarm users might be using a device with SD/eMMC support, they will encounter the bug and that needs to be fixed.

### DIFF
--- a/core/systemd/PKGBUILD
+++ b/core/systemd/PKGBUILD
@@ -13,8 +13,8 @@ pkgname=('systemd'
          'systemd-resolvconf'
          'systemd-sysvcompat'
          'systemd-ukify')
-_tag='34818a93c640ce377f0d00849a74e029624b4a12' # git rev-parse v${_tag_name}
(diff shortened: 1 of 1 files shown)


7Ji:
Close as already updated in 01d90d56f7cf5153ea8380c11e381496552d7c51

PR #2009, opened 2023-09-27
Re-enable qemu-user-static builds

ewlsh:
Static builds were failing due to incorrect build flags in glibc 2.35.

See https://sourceware.org/bugzilla/show_bug.cgi?id=29514.

I'm currently compiling this locally with a patched glibc 2.35 but glibc 2.38 is in this repository which should compile correctly unpatched.

I removed x86-only dependencies from `qemu-system-x86` (seabios, edk2-ovmf) based on what I saw [in Fedora's packaging](https://src.fedoraproject.org/rpms/qemu/blob/rawhide/f/qemu.spec) but if these are compilable on aarch64 let me know and I can adjust.

(my local patch for glibc 2.35)

```
diff --unified --recursive --text a/Makeconfig b/Makeconfig
--- a/Makeconfig        2023-09-24 23:08:32.340839948 -0700
+++ b/Makeconfig        2023-09-24 23:11:54.619347519 -0700
@@ -1070,7 +1070,7 @@
 PIC-ccflag = -fPIC
 endif
 # This can be changed by a sysdep makefile
-pie-ccflag = -fpie
+pie-ccflag = -fPIE
 no-pie-ccflag = -fno-pie
 # This one should always stay like this unless there is a very good reason.
 PIE-ccflag = -fPIE
diff --unified --recursive --text a/sysdeps/sparc/Makefile b/sysdeps/sparc/Makefile
--- a/sysdeps/sparc/Makefile    2023-09-24 23:08:32.552836773 -0700
+++ b/sysdeps/sparc/Makefile    2023-09-24 23:16:15.058325731 -0700
@@ -1,9 +1,6 @@
 # The Sparc `long double' is a distinct type we support.
 long-double-fcts = yes
 
-pie-ccflag = -fPIE
-no-pie-ccflag = -fno-PIE
-
 ifeq ($(subdir),gmon)
 sysdep_routines += sparc-mcount
 endif
 ```

### DIFF
--- a/extra/qemu/PKGBUILD
+++ b/extra/qemu/PKGBUILD
@@ -5,7 +5,7 @@
 # ALARM: Kevin Mihelich <kevin@archlinuxarm.org>
 #  - remove makedepends on seabios, revert not installing qemu seabios files
 # ALARM: Qu Wenruo <wqu@suse.com>
-#  - remove static build, make qemu-base/qemu-desktop meta packages to use their native system emulator as default dependency
+#  - make qemu-base/qemu-desktop meta packages to use their native system emulator as default dependency
 
 pkgbase=qemu
 pkgname=(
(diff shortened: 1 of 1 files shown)


ewlsh:
If glibc 2.38 can't be released yet (for whatever reason), let me know if there is a way to propose a patch to glibc 2.35 to unblock this :sun_with_face:

SpieringsAE:
Can this be done? I want set up a archlinuxriscv chroot on my device and it needs this.
I just compiled it myself and that seems to have worked so far so I don't see why not.

PR #2011, opened 2023-10-19
Expose ability to control fan speed via sysfs

millw:
Enable Kernel configuration to allow sysfs fan control.


graysky2:
I do not see this option listed in the `make nconfig` output

PR #2023, opened 2024-03-09
Adding aarch64 for extra/netdata

mys721tx:
Adds aarch64 for extra/netdata. It might have caused a problem with the missing go plugin.


graysky2:
This package is already built for aarch64.

PR #2030, opened 2024-04-11
core/linux-rpi setting CONFIG_ARM64_VA_BITS_48 for better software compat

lucasoskorep:
This PR aims to update the linux-rpi kernel to use CONFIG_ARM64_VA_BITS_48 instead of CONFIG_ARM64_VA_BITS_39.

Unfortunately there is quite a lot of software (especially in the server hosting world) that is written for arm assuming VA_BITS_48 is set.  48 is backwards compatible with 39 so this patch will only increase software compatibility with the linux-rpi kernel.  The only downside is that there is a nominal increase in memory usage when using 48 instead of 39. 

This change also affects a small handful of other dependant variables including:
- CONFIG_ARCH_MMAP_RND_BITS_MAX
- CONFIG_PGTABLE_LEVELS
- CONFIG_GCC_ASM_GOTO_OUTPUT_WORKAROUND
- CONFIG_GCC10_NO_ARRAY_BOUNDS

but these values should all be automatically derived from the main CONFIG_ARM64_VA_BITS_48.

For context this kernel change has been made by many other rpi linux distros including fedora, debian, and ubuntu.

There is some further context/discussion around this issue on the raspberry pi issue tracker https://github.com/raspberrypi/linux/issues/4375

If this solution is not acceptable I would like to discuss the possibility of creating a new  linux-rpi-server package which has this patch and any others needed for general stability increases for alarm on the raspberry pi.


graysky2:
I don't fully understand the implications of making this change.  From reading the issue you linked, it doesn't seem that upstream is sold on the idea.

lucasoskorep:
> I don't fully understand the implications of making this change. From reading the issue you linked, it doesn't seem that upstream is sold on the idea.

So CONFIG_ARM64_VA_BITS_48 is for setting the size of the the virtual address space, making it 48 bits instead of 39 bits.  

Taking the numbers from the rpi kernel team, this can use up to 64KB per page table, but on average an additional 30KB per page table.  Roughly 3mb per 100 processes and a fresh install has ~200 processes running so 6 more megabytes of ram used. 

CONFIG_PGTABLE_LEVELS-4 means there is one more virtual page table level to go through and the others are just adjustments to branching workarounds.  

ARM software being developed for server sided applications is generally targeting cloud arm environments which have universally adopted the same standard of 4KB page size, 48bit VA bits, and 4 levels of page table redirection for the sake of standardization.  While software could check for each of these requirements, there is a non-insignificant portion that does not and refuses to run on non-conforming software.

SO that being the case, most non-raspbian arm distributions for the pi3, pi4, and pi5 have all adopted these settings as standard in their kenel.  These distros being, fedora, ubuntu, debian, armbian, apertis, and im sure others (although I believe debian and armbian share some repos and possibly their kernel)

If you/alarm wants to keep this change out to preserve a bit of memory consumption on older boards such as the pi 3 that works for me, but I would like to help in maintaining a separate kernel for the rpi then that includes this configuration to allow for maximum software compatibility for people who want it.

graysky2:
If the numbers you're quoting are accurate, I don't see that as much of a problem but what is the increased compatibility trade off?  What are use-cases for this change and do they affect enough Arch ARM users is what I am thinking...

lucasoskorep:
> If the numbers you're quoting are accurate, I don't see that as much of a problem but what is the increased compatibility trade off? What are use-cases for this change and do they affect enough Arch ARM users is what I am thinking...

Im not sure exactly what kinds of cases I can make here as I can only speak for myself but I have found 5 independant programs I have tried to run have been broken due to this kernel issue. 

From a more general standpoign - software written using  google's TCMalloc appears to have issues when using default configurations -  https://github.com/google/tcmalloc. 

Personally I could not run [stackgres,](https://stackgres.io/) [postgres-operator,](https://github.com/zalando/postgres-operator) and [istio](https://istio.io/),  and a project istio is based off of - [envoy](https://github.com/envoyproxy/envoy/issues/15235#issuecomment-850516622).  The first two are kubernetes database operators and are somewhat niche. 

The latter 2 are the bread and butter of many kubernetes and other container based workloads. 

Also OpenJDK compiled projects can have issues as many versions of openjdk do not support CONFIG_PGTABLE_LEVELS=3 which is a downstream affect of 39bit virtual memory space. 

https://bugs.openjdk.org/browse/JDK-8252500?page=com.atlassian.jira.plugin.system.issuetabpanels%3Aall-tabpanel

Cloudflare's wrangler CLI for interacting with cloudflare workloads has issues, as well as other parts ofh te cloudflare CLI - https://developers.cloudflare.com/workers/wrangler/

Also anything compiled using [cosmopolitan](https://github.com/jart/cosmopolitan) is current broken and doesnt neet ot be. 

Overall this can affect container workloads as well as even just normal java projects fairly easily.  The prevalence of this is hard to exactly pin down but it is fairly wide spread across different languages and stacks.

PR #2032, opened 2024-06-06
[thunderbird] Fix missing stdsimd on ARM

mogwa1:
This fixes the rust errors during build.


mogwa1:
I ran into some issues at the end of the build.  I had to use `nodejs-lts-iron` (=v20) instead of `nodejs` (=v22) to get a successful build.  I'm still looking into the underlying reasons for this.

mogwa1:
Closing since it seems to be fixed in 115.12.2 now.
I still need `nodejs-lts-iron` though...

PR #2037, opened 2024-06-28
linux-aarch64: Enable Support For SC8280XP

theironrobin:
I noticed a few issues booting this kernel on the ThinkPad X13s. Setting these configs should help.


graysky2:
Please squash into a single commit and bump the `pkgrel`, see: https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md

Also, do your proposed changes boot and run on the target hardware (ie tested and verified by you)?

theironrobin:
no, looking like this still isn't working.

PR #2043, opened 2024-07-08
core/linux-aarch64: enable smb server

TheDcoder:
The Arch Linux kernel has had this [enabled](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/commit/62812fc5b459a08f1366b1cd46eb379b3d37b26f) for a while now.

I requested this feature [over a year ago](https://archlinuxarm.org/forum/viewtopic.php?f=65&t=16592&p=71556) on the forum as well but I was surprised to see that it still wasn't added!

Also it might be a good idea to package [`ksmbd-tools`](https://aur.archlinux.org/packages/ksmbd-tools) in the `aur` repository :smile:


graysky2:
Please see my comment, update the checksums, and bump the pkgrel in your PR to make it suitable for merging.

PR #2046, opened 2024-07-17
core/linux-odroid-xu3: Add conflicts/replaces for linux-odroid-n2

polyzen:
"Yes, we are using odroid-6.6.y for C4 Series, N2 Series and XU4."
https://github.com/hardkernel/linux/issues/437#issuecomment-2233498714


polyzen:
Rebased and adjusted for eca2a8eb47803ef9637a29acf66f922376f6d216

polyzen:
Did not realize the C4 and N2 were arm64. The linux-odroid-n2 pkgbuild would need to be revamped, but I do not have either of those devices with which to test on.

PR #2047, opened 2024-07-17
alarm/uboot-sunxi to 2024.07-2

linkmauve:
makedepends was missing python-setuptools.

Fixes build broken by #2045.

### DIFF
--- a/alarm/uboot-sunxi/PKGBUILD
+++ b/alarm/uboot-sunxi/PKGBUILD
@@ -18,11 +18,11 @@ pkgname=('uboot-a10-olinuxino-lime'
          'uboot-pcduino3'
          'uboot-pcduino3-nano')
 pkgver=2024.07
-pkgrel=1
+pkgrel=2
 arch=('armv7h')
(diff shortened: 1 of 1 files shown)


graysky2:
@linkmauve - do you have hardware that requires this package and if so, can you please confirm that this PR is functional on that hardware by adding the following to your commit message:

```
Run-tested: <hardware model>
```

ejona86:
I was able to build the package in a clean chroot with this change and it successfully booted on an A20-Olinuxino-Lime2 via an SD card (I did not try other boot methods).

linkmauve:
Aaaah sorry, I had tested them but forgot to reply afterwards…
```
Run-tested: A10-Olinuxino-Lime
Run-tested: A20-Olinuxino-Lime
Run-tested: A20-Olinuxino-Lime2
```

Out of ten boots each, they all booted fine (unlike the previous 2017.01 version where the Lime2 wouldn’t boot half of the time), and gigabit ethernet also works on the Lime2 (tested at about 30 MiB/s on the LAN).

PR #2049, opened 2024-07-29
linux-armv7: chromebook: Actually ship device tree binaries

urjaman:
This is untested (i just did s/dts/dtb/ in the github editor), but just saying that trying to boot my fork with the .dts (just a fragment of source...) in place of a .dtb didnt work.

(Also makes me wonder whether you tested the chromebook boot ...)


graysky2:
Please don't submit untested PRs.  Do you have this hardware and can you test your proposed change and verify that it works?

urjaman:
Are you trying to say that you don't have the hardware and would like a tester for linux-armv7-chromebook?  

If so, please work on the phrasing... because it feels like holier than thou to ask me to test the fix (on linux-armv7 - i did test it, just elsewhere) when you yourself shipped an untested change to the world (that I copied a part of, and wasted my time).

Anyways, my point was to just show that there is an issue and suggest a fix at the same time (since issues are disabled, a PR was the only way.).

What I can say is that https://github.com/urjaman/linux-armlfs/blob/main/kernel.its <- that works.

graysky2:
I do not have a chromebook for testing and asked if you did.

urjaman:
Okay, in that case, whilst testing linux-armv7 i found out that I needed to also fix your config ;)

urjaman:
So, once you _do_ get the testing (and an extra fix too) you requested (=I spent my time), it's just silence; great.

I'll get back to just making a super low-effort PR if I ever feel the need to report an issue.

graysky2:
I am confused. You did test the PR on the hardware? I was waiting to hear that you confirmed function to take an action.

graysky2:
Rebasing this against master will be simple but I need to know that the changes have been properly tested.  Can you confirm that the PR as submitted works on the hardware as well as list the hardware on which you tested please?

urjaman:
The PR (as of my force-pushed update 2 weeks ago) boots with depthcharge (original chromebook fw) on an ASUS C201 chromebook (aka veyron speedy).

graysky2:
Thank you for confirming functionality and for contributing.

PR #2050, opened 2024-07-29
Add support for bcachefs in linux-aarch64

nnathan:
First time contirbuting.

I would like to see support added for bcachefs.

It is marked experimental in the kernel, but it's been included in archlinux `linux` package [since its inception in the kernel](https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/blob/79425d7496066714f38535eb4f37df529668e83a/config#L10380-10388).

I've followed contributing guidelines such that I've confirmed it builds in a clean chroot. Though I was confused as to whether I should added a `.1` to pkgrel or bump it, I went with the latter since this doesn't track an upstream x86 arch linux package.


nnathan:
Update to 6.10.4 kernel, can someone please take a look at this PR? Thanks.

nnathan:
Thanks heaps @graysky2. Since I don't have a way to contact the people that maintain the website, maybe you can assist. The installation instructions for alarm on rasp pi 4 is incorrect. My [gist](https://gist.github.com/nnathan/1ab33fd4b7518b1ba97a28e99fcbe2cd) covers how to install on a Rasp Pi 4 but it bcachefs flavoured. But particularly step 10 and and the _note_ after step 14 needs to be adhered to install based on [these instructions](https://archlinuxarm.org/platforms/armv8/broadcom/raspberry-pi-4).

PR #2053, opened 2024-08-08
Patch to prevent freeze during boot if gpio device connected

RononDex:
This fixes an issue where u-boot would hang whenever a device (for example the GPS hat from Adafruit)  is connected to the raspberry pi's GPIO header.

These devices usually output something to the serial console during the startup phase of u-boot, making u-boot think it has to interrupt its boot sequence because somebody is trying to interrupt it's startup.

This also aligns the rpi4 behavior with the one from Raspbian, which does not interrupt its boot sequence with a GPIO device attached


graysky2:
Is there a reference to the included patch?  Please bump the `pkgrel`

RononDex:
The patch is copied from here: https://raspberrypi.stackexchange.com/questions/116074/how-can-i-disable-the-serial-console-on-distributions-that-use-u-boot

Pkgrel is now raised to 4

graysky2:
Merged thank you for the PR

PR #2057, opened 2024-08-27
Apply patch from raspberry kernel to fix PPS signal on 64-bit ARM

RononDex:
This fixes PPS read timeouts on 64-bit ARM caused by another bug fix in the upstream kernel that was aimed for 32-bit ARM but was wrongly also applied to 64-bit.

See https://github.com/raspberrypi/linux/issues/5430 for further information. Patch is also taken from there


graysky2:
Thank you for the PR.  In the future, please review [CONTRIBUTING.md](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md).  One PR per package.  You also accidently bumped the rc version to 6 rather than the pkgrel.

RononDex:
Will do and thanks :)

PR #2064, opened 2024-09-19
enable CONFIG_SECURITY_LANDLOCK...

eworm-de:
... now that pacman can make use of if in version 7.0.0 when setting this option in /etc/pacman.conf:

DownloadUser = alpm


eworm-de:
I guess this should be changed in all Linux packages...

graysky2:
Indeed, thanks for the PR.

PR #2065, opened 2024-09-25
newpkg: extra/intellij-idea-community-edition

TheOneWithTheBraid:
Tested on :

- `aarch64`: Qualcomm Snapdragon 8cx (Microsoft Volterra and Lenovo ThinkPad X13s)


TheOneWithTheBraid:
Streamlined with https://gitlab.archlinux.org/archlinux/packaging/packages/intellij-idea-community-edition/-/merge_requests/4 - IntelliJ by default cross-compiles to any supported architecture, I explicitly disabled this as a contribution to the upstream Arch Linux package.

PR #2068, opened 2024-11-02
Add `uboot-radxa-zero-3` package

vially:
Add `u-boot` package for Radxa [Zero 3E](https://radxa.com/products/zeros/zero3e/)/[3W](https://radxa.com/products/zeros/zero3w/) single board computers.

The package has been successfully tested on both SBCs (using the generic [linux-aarch64](https://archlinuxarm.org/packages/aarch64/linux-aarch64) kernel package) and it seems to work as expected.


vially:
Is this package something the ArchLinux ARM maintainers are interested in? I have a few more similar packages that I would like to submit for inclusion into this repository, but it's not clear whether it's preferred to add them to this repository or if I should just keep them in the AUR.

PR #2080, opened 2025-01-04
linux-aarch64: remove unused `CONFIG_USER_NS_UNPRIVILEGED` config

vially:
It [turns out](https://archlinuxarm.org/forum/viewtopic.php?f=23&t=14193) this config flag is not part of mainline so it doesn't actually do anything on Arch Linux ARM.

For anyone else stumbling on this while trying to get rootless podman working on Arch Linux ARM, the workaround is to [add extra capabilities](https://github.com/containers/podman/issues/2788#issuecomment-479972943) to the `newuidmap`/`newgidmap` binaries:
```sh
$ sudo setcap cap_setuid+ep /usr/bin/newuidmap
$ sudo setcap cap_setgid+ep /usr/bin/newgidmap
```


graysky2:
Thank you for the PR... I caught this earlier but did not push the commit, see: https://github.com/archlinuxarm/PKGBUILDs/commit/11b2ee27f90b326964b8a19ac4a76a79fd938986

PR #2081, opened 2025-01-04
core/linux-aarch64: add support for Turris Mox rWTM

bschnei:
The [Turris Mox](https://www.turris.com/en/products/mox/) uses the Marvell Armada 37xx chipset which is the same chipset as that used in [ESPRESSObin](https://archlinuxarm.org/platforms/armv8/marvell/espressobin). These devices contain an internal Cortex-M3 processor that provides hardware RNG. Compiling support for this hardware as a kernel module should allow this generic kernel package to support the ESPRESSObin. This device likely does not need its own [device-specific kernel](https://github.com/archlinuxarm/PKGBUILDs/tree/master/core/linux-espressobin).


graysky2:
Do you have this hardware and can verify functionality with this module?

bschnei:
Hi @graysky2, yes, I own two ESPRSESObin Ultra's which have the same Aramda 37xx chipset and have been running kernels with this option compiled as a module for about a year. The module will only load if the device's firmware exposes the rWTM which will only happen if the bootloader is built to support it. Both of my devices load this module and the kernel provides `/dev/hwrng` to userspace. Without this module, there is no `/dev/hwrng` available.

I've been working on the [bootloader](https://github.com/bschnei/ebu-bootloader) and [kernel](https://github.com/bschnei/linux-a3700) for this device for about a year now and would be happy to help support A37xx devices here. Sadly, the devices as shipped from Globalscale have a lot of issues with the bootloader that will make them difficult to support. In particular, UEFI booting is broken when using Linux kernels > 6.0.

PR #2087, opened 2025-01-27
Blender update

yuyichao:
This includes adding a few packages that are probably holding back the latest blender from being built.
The usd package was also in this dependency chain but it should build just fine using the upstream PKGBUILD as long as its dependencies `embree` and `materialx` are built.


scriptforfivem:
Your approach to this problem is very elegant. I'm learning a lot from reviewing your code.

PR #2089, opened 2025-02-09
add package: extra/uv

sapslaj:
Similar to some other packages, uv needs to be built with `JEMALLOC_SYS_WITH_LG_PAGE=16` to work on some aarch64 systems. Upstream is doing this for their official builds (https://github.com/astral-sh/uv/issues/6528) and I confirmed this works on my Apple M1. The PKGBUILD from official Arch doesn't set this, so the `uv` package is broken and crashes. This change should fix that. The rest of the additions are just a copy from Arch's packaging repo.

### DIFF
--- a/extra/uv/PKGBUILD
+++ b/extra/uv/PKGBUILD
@@ -1,0 +1,90 @@
+# Maintainer: Orhun Parmaksız <orhun@archlinux.org>
+# Maintainer: Caleb Maclennan <caleb@alerque.com>
+# Contributor: MithicSpirit <rpc01234 at gmail dot com>
+# Contributor: David Runge <dvzrv@archlinux.org>
+# Contributor: Leonidas Spyropoulos <artafinde@archlinux.org>
+# Contributor: Daniel M. Capella <polyzen@archlinux.org>
+# Contributor: Guillaume Gauvrit <guillaume@gauvr.it>
+
+# ALARM:
+#  - build aarch64 with 16k page support
+
+pkgbase=uv
+pkgname=("$pkgbase" "python-$pkgbase")
+pkgver=0.5.29
+pkgrel=1.1
+pkgdesc='An extremely fast Python package installer and resolver written in Rust'
+arch=('x86_64')
+url="https://github.com/astral-sh/uv"
+license=('MIT' 'Apache-2.0')
+depends=('gcc-libs' # 'libgcc_s.so'
+         'glibc' # 'libc.so' 'libm.so'
+         'zlib' # 'libz.so'
+         'bzip2' # 'libbz2.so'
+         )
+makedepends=('cargo' 'maturin' 'python-installer' 'cmake' 'git')
+checkdepends=('python' 'python-zstandard' 'libxcrypt-compat' 'clang')
+options=('!lto')
+source=("git+$url.git#tag=$pkgver")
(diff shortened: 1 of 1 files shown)


graysky2:
We don't need the `.files` only the PKGBUILD. Also, see [here](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#upstream-x86-arch-linux-packages) regarding the pkgver. Should be 1.1 not 2. I guess building with this flag for aarch64 means that that flag is compatible with all aarch64 regardless of page sizes?

sapslaj:
> We don't need the `.files` only the `PKGBUILD`.

I figured as much, but wasn't sure since other packages seem to include those.

> Also, see [here](https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#upstream-x86-arch-linux-packages) regarding the pkgver. Should be 1.1 not 2.

Ah yes, completely misinterpreted that part. My bad!

> I guess building with this flag for aarch64 means that the resulting package is 100% compatible with all aarch64 hardware regardless of page sizes?

Yep. 16K is inherently 4K aligned so it will work fine on any of those machines. I think _technically_ 64K page sizes is "safer" but in practice 16K will most likely be totally fine as I'm not aware of any aarch64 machines that need 64K page sizes. For what it's worth, there's already precedent for this with the [jemalloc](https://github.com/archlinuxarm/PKGBUILDs/blob/master/extra/jemalloc/PKGBUILD#L31), [fd](https://github.com/archlinuxarm/PKGBUILDs/blob/2a4ee8c6aa8e8de34f0dcabeed8aa85190dd2b8b/extra/fd/PKGBUILD#L28), and [rust](https://github.com/archlinuxarm/PKGBUILDs/blob/2a4ee8c6aa8e8de34f0dcabeed8aa85190dd2b8b/extra/rust/PKGBUILD#L208) packages. Also, like I mentioned, this is what official builds of uv do so it seems reasonable to me to follow in their footsteps. This issue seems to be pretty prevalent since there's a [whole section on the Asahi Linux docs about it](https://github.com/AsahiLinux/docs/wiki/Broken-Software#why-not-just-use-4k-pages-then) (with links to [fixes in this repo](https://github.com/archlinuxarm/PKGBUILDs/commit/19a1393) as well as a [link to the upstream jemalloc issue](https://github.com/jemalloc/jemalloc/issues/467)). I'm no expert though. @marcan and/or @kmihelich might have more insight.

In the meantime I'll get the other two things fixed and amend my commit.

sapslaj:
Alright, change pushed. Let me know if you see anything else that needs addressing @graysky2. And thanks for the review!

sapslaj:
Also, looks like we'll have to do the same tweak for [ruff](https://archlinux.org/packages/extra/x86_64/ruff/) after this one.

```
❯ getconf PAGESIZE
16384

❯ ruff --version
<jemalloc>: Unsupported system page size
<jemalloc>: Unsupported system page size
memory allocation of 48 bytes failed
fish: Job 1, 'ruff --version' terminated by signal SIGABRT (Abort)
```

graysky2:
> Also, looks like we'll have to do the same tweak for [ruff](https://archlinux.org/packages/extra/x86_64/ruff/) after this one.

I don't use either of these packages, but if merging this breaks the other, we want to get in front of that before the merge.

graysky2:
Thanks for clarifying. Plz do.

PR #2092, opened 2025-03-03
core/linux-rpi: make boot option specific for pi4...

eworm-de:
... as Raspberry Pi 4 requires this, but Raspberry Pi 2 must not have it.

### DIFF
--- a/core/linux-rpi/PKGBUILD
+++ b/core/linux-rpi/PKGBUILD
@@ -32,7 +32,7 @@ source=("linux-$pkgver-${_commit:0:10}.tar.gz::https://github.com/raspberrypi/li
 )
 md5sums=('3315e570d75d48a2f3c2de336510585e'
          '3bab7426d8c8818dda8353da3892a41f'
-         '44f69a0637b34eb772980614f9246d01'
+         '577f0f85bd08490fdd8ebe47d2486463'
          'a157c5bfc0f03d0728c92bd953b06265'
          '8b93b6ca167f70c60277f352f7b78024'
          '4d7ab0ab2625470ebb916176b5d9cf02')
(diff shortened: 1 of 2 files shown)


eworm-de:
Though not sure about Raspberry Pi 3 models... Do they need it?
Let me know... Happy to updates this.

eworm-de:
But this package is built for `armv7h`, so RPi2 can run it. No?

graysky2:
Yes, but that argues for two config.txt files depending on the CARCH

eworm-de:
Ah, I missed that the package is built for two architectures. Anyway, I would go with one file, as conditional filters (which do a fall-through) make this easy and straight forward. Should work this way, no?

Let me know if you still prefer two files, I will update again then.

eworm-de:
Oh, this package does not have a `.SRCINFO` file? Just dropped that from my commit.

graysky2:
Where is your reference that pi2 cannot boot with these lines? Here is [config.txt](https://github.com/RPi-Distro/pi-gen/blob/master/stage1/00-boot-files/files/config.txt) from RPiOS I believe.

eworm-de:
Well, it does boot, but `xorg` fails to display anything. The display is just black.

I searched for something like "*raspberry pi 2 black screen xorg*", and found several people having the same issue. One of these results gave a hint to drop the given setting from `config.txt`. I should have recorded the url, can not find it now. 🤨

After all still (or increasingly) unsure myself... But it works for me.

graysky2:
OK so you have a RPi2 and if you boot with those lines disabled you get the expected behavior. If you add those lines back and reboot, you have no video output?

eworm-de:
Exactly. (That goes for xorg, and tty is fine any way.)

PR #2093, opened 2025-03-03
Enable binder IPC & binderfs

yaroslav-ilin:
Porting over the changeset from x86_64 Arch Linux kernel. This module is built-in in upstream since kernel `6.12.7-1`.

From the upstream merge request:

> [BinderFS](https://www.kernel.org/doc/html/latest/admin-guide/binderfs.html), which is necessary for running [Waydroid](https://wiki.archlinux.org/title/Waydroid), has been in upstream Linux kernel for a long time. In [Fedora](https://discussion.fedoraproject.org/t/discussion-about-waydroid-especially-the-kernel/36991/5), standard Linux kernel has included binderfs since v5.15. Therefore, it would be better for Arch Linux to enable binderfs.

https://gitlab.archlinux.org/archlinux/packaging/packages/linux/-/merge_requests/5


graysky2:
Thanks for the PR, adjusted pkgver

PR #2097, opened 2025-03-18
core/linux-aarch64-rc: Enable more Qualcomm X Elite config options

andersson:
The recently merged patch to enable Qualcomm X Elite was missing two
drivers in the USB2 path, Qualcomm UCSI driver and the audio subsystem
pinctrl driver. Enable these.

Also set the newly introduced CONFIG_BT_HCIBTUSB_AUTO_ISOC_ALT to its
default value.


graysky2:
Yes, I posted to the wrong one.

PR #2098, opened 2025-03-18
core/linux-aarch64: Enable more Qualcomm X Elite config options

andersson:
The recently merged patch to enable Qualcomm X Elite was missing two
drivers in the USB2 path, Qualcomm UCSI driver and the audio subsystem
pinctrl driver. Enable these.

To fix up the config, CONFIG_LAN966X_OIC is disabled per its default,
and in line with its dependency on CONFIG_MCHP_LAN966X_PCI in the
upcoming v6.15-rc1.


graysky2:
This config seems to be missing:
```
# CONFIG_LAN966X_OIC is not set
```
Which should be inserted after 
`CONFIG_HISILICON_IRQ_MBIGEN=y`

andersson:
I'm sorry, but as far as I can tell it's there. When I open the commit associated with this PR it does include the line you're referring to, and it's described in the commit message.

Then looking at the other PR (for the -rc kernel) it's not there, because in v6.14-rc this now has a dependency on another config option that we don't have enabled.

graysky2:
Merged in https://patch-diff.githubusercontent.com/raw/archlinuxarm/PKGBUILDs/pull/2097.patch

PR #2102, opened 2025-04-18
extra/chromium: fix

mkurz:
Same fix which was just merged upstream: https://gitlab.archlinux.org/archlinux/packaging/packages/chromium/-/merge_requests/8


mkurz:
Already in master now.

PR #2103, opened 2025-04-24
Delete old and unmaintained kernels

solskogen:


kmihelich:
Old/unmaintained doesn't mean not still in use, which some of these are.

solskogen:
There's been over 1,600 CVEs reported since 2021, which these kernel might be exposed to. I don't think it's a good idea to have them.

PR #2105, opened 2025-04-25
Delete core/file directory

solskogen:
seccomp is available on both armv7 and aarch64, so there's no reason to have it any other way than on upstream.


kmihelich:
This wasn't about seccomp availability.  This is here because of #1757, and you've checked this?

PR #2106, opened 2025-04-25
Delete core/icu directory

solskogen:
the workarounds are not needed anymore


kmihelich:
Debian is still using that patch.  What have you done to show it's no longer necessary?

solskogen:
Checked config.log if it was picking up distcc and/or clang during building. I was just using icu from upstream.

kmihelich:
I'm not referring to the distcc workaround.

solskogen:
Ah, but Arch doesn't include it (and at first glance it doesn't look like they ever did), and if that workaround is really needed my initial thought was that it should be there and not here. The issue some had in 2013 with problems liking seems to gone on my round of testing.

kmihelich:
The patch is still required for v7.

PR #2108, opened 2025-04-25
Delete core/openssl-cryptodev directory

solskogen:
Removed from upstream, and nothing depends on it.


kmihelich:
This wasn't an upstream package, but it's not being kept up to date anymore.

PR #2116, opened 2025-05-14
Depend on initramfs instead of mkinitcpio

clarfonthey:
This removes the `mkinitcpio` dependency from most kernel packages and replaces it with `initramfs` to allow alternative initcpio creation, which is done on mainline Arch.

Small note: these depend on `mkinitcpio>=0.7`, but `mkinitcpio` only started providing `initramfs` after this version, so, this replacement is valid.

The following kernel packages were *excluded* from this change because they provide custom hooks that run `mkinitcpio` directly:

* `linux-am33x`
* `linux-armv7-rc`
* `linux-espressobin`
* `linux-odroid-c2`
* `linux-odroid-n2`

These hooks are probably unnecessary because `mkinticpio` provides its own hook, but I didn't want to make any invasive changes.

### DIFF
--- a/core/linux-aarch64-rc/PKGBUILD
+++ b/core/linux-aarch64-rc/PKGBUILD
@@ -11,7 +11,7 @@ _srcname=linux-${_rcver}-rc${_rcrel}
 _kernelname=${pkgbase#linux}
 _desc="AArch64 multi-platform (release candidate)"
 pkgver=${_rcver}.rc${_rcrel}
-pkgrel=1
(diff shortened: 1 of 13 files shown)


clarfonthey:
Note: I am more than happy to rebase this once someone reviews it, but I'm not going to constantly rebase it until then. I assume that all these kernel package updates are automated.

clarfonthey:
Not worth keeping this open if nobody's going to merge it.

PR #2126, opened 2025-06-20
core/linux-aarch64-rc: Qualcomm X Elite updates for v6.16

andersson:
The Synopsys eUSB2 driver config changed in v6.16, select the new option to ensure USB still works on a variety of Qualcomm platforms.

X Elite CPU frequency scaling was enabled in v6.16, this depends on the the Qualcomm CPUCP mailbox driver, so enable this.

Support for external display was added to a number of X Elite-based boards, depending on the PS883X retimer driver, enable this.

Provide olddefconfig value for CONFIG_RUST to keep config in sync.


graysky2:
TY

PR #2129, opened 2025-06-26
Delete extra/gnome-power-manager directory

solskogen:
compiles fine on aarch64


kmihelich:
Patch is still required.

PR #2134, opened 2025-07-25
core/linux-odroid-xu3 to 6.6.93-1

polyzen:
Patches against the odroid-6.6.y branch taken from Armbian:
https://github.com/armbian/build/tree/main/patch/kernel/archive/odroidxu4-6.6


graysky2:
I have none of these SoCs. You're saying that the XU4 runs this kernel package and works fine? Do we have any affected hardware to confirm functionality?

PR #2139, opened 2025-08-11
core/linux-aarch64: enable Synopsys `hdmirx` module

vially:
Enable the Synopsys HDMI-RX module that is used by various Rockchip RK3588 single board computers (e.g.: Radxa ROCK 5B).


graysky2:
Plz bump pkgrel

PR #2151, opened 2025-10-24
core/pacman: fix $CFLAGS

P-Daddy:
The `-mno-omit-leaf-frame-pointer` flag isn't valid on ARMv7.

Testing done implicitly by nature of the fact that I have a patch for a different `PKGBUILD` to _remove_ this flag from `$CFLAGS` and `$CXXFLAGS` to resolve an error from `gcc`.


kmihelich:
Fixed in 3904764a6f6b3a8d2fa66c1b87d8b26172dd55c2

PR #2191, opened 2026-03-30
added alarm/stubble-git

abelvesa:
Stubble is a minimal UEFI kernel boot stub that generates hwids of the 
running machine derived from smbios, compares them to an embedded 
lookup table in the .hwids section of the kernel image, and then selects
the machine specific devicetree blob from the .dtbauto section of the
same kernel image.


andersson:
Would it not be preferable to have the problem statement of the PR in the git message?

Also, what about S-o-b?

PR #2194, opened 2026-04-27
core/linux-aarch64-rc to v7.1.rc1-1

andersson:
Bump linux-aarch64-rc to v7.1-rc1 and update "config" to olddefconfig


graysky2:
Thanks for the PR

PR #2196, opened 2026-05-01
Update AM33xx kernel to 7.0

davidjohnsummers:
This patch updates the kernel config to 7.0

Not this does not include cfg80211 - which is needed for wireless. This is because compiling this takes over 0.5GB or memory, and on a beagle this triggers the OOM killer. A way round this has not been found.


graysky2:
Swap file?

davidjohnsummers:
I never set up a partition for a swap file, and now would need to reformat the SD card that I don't particular want to do. Anyway with gcc, I suspect it would access its memory in a random direction, which means it would need to keep on accessing the spaw file.

graysky2:
You do not need a dedicated partition: https://wiki.archlinux.org/title/Swap#Swap_file

PR #2198, opened 2026-05-04
Linux kernel 7.0 for am35xx devices

davidjohnsummers:
This adds the missing package.

It updates the linux kernel for am35xx devices to the 7.0 kernel.

The modules have been compressed (xz) to save space
The firmware has been removed as a dependancy - as the am35xx chips have no firmware in that package
The firmware for the M3 processor is now included in the package
The device tree have been updated to the current valid versions

This has been tested on both a BBB and on a pocketbeagle.

### DIFF
--- a/core/linux-aarch64/PKGBUILD
+++ b/core/linux-aarch64/PKGBUILD
@@ -8,7 +8,7 @@ pkgbase=linux-aarch64
 _srcname=linux-7.0
 _kernelname=${pkgbase#linux}
 _desc="AArch64 multi-platform"
-pkgver=7.0.4
+pkgver=7.0.5
 pkgrel=1
 arch=('aarch64')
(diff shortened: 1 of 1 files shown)


graysky2:
Can you plz rebase squash into a single commit: https://github.com/archlinuxarm/PKGBUILDs/blob/master/CONTRIBUTING.md#rules-for-pull-requests

davidjohnsummers:
Will do when I get time. Most of the submit where going through many kernel changes, one commit from each kernel - you ignore most, now I'll need to back those out and do it all in one go.

The only commit that was different was the one that added linux-am33x-7.0-1-armv7h.pkg.tar.xz - that was over 25MB, and couldn't upload with the www to github as to large. Only noticed later it hadn't uploaded - and so pushed it from my machine.

PR #2201, opened 2026-05-19
core/linux-odroid-xu3 to 6.6.140-1

polyzen:


graysky2:
Thanks for the PR, merged but had to update the checksums

PR #2202, opened 2026-05-30
linux-aarch64: enable MMC_SDHCI_BRCMSTB

PolynomialDivision:
```
The Raspberry Pi 5 (BCM2712) SD card controller declares itself as:
  compatible = "brcm,bcm2712-sdhci", "brcm,sdhci-brcmstb"

CONFIG_MMC_SDHCI_BRCMSTB was explicitly disabled, so the kernel has no
matching driver. Without it the SD card is invisible to the kernel and
every Pi 5 boot ends with:

  timeout waiting for device /dev/disk/by-partuuid/...
  Dependency failed for /sysroot
  (drops to emergency mode)

Enabling this option compiles the sdhci-brcmstb driver into the kernel
and fixes Pi 5 boot from SD card. Tested on Raspberry Pi 5.
```


graysky2:
Thanks, fixed-up a little in 96482856a9aa9ef4be3e399881c435c6d07e0e5f

PR #2204, opened 2026-06-11
linux-aarch64-rc: enable configs for ASUS Zenbook A14 (UX3407QA)

aotot:
The X1P42100 GPU clock controller and the WSA884x speaker amp codec are off in the config. This commit changes that.

This has been tested on real hardware. A lot (nearly everything) works. Note that one still needs to follow https://github.com/alexVinarskis/linux-x1e80100-zenbook-a14 to get the necessary firmware.


graysky2:
thx

PR #2207, opened 2026-06-20
Update linux kernel for am335x devices to 7.1

davidjohnsummers:
This is a single monolithic update from the current arch linux arm 6.13 version to the 7.1 version. This has been done sequentially previously through something like 5 updates that were not accepted by arch linux arm, as so were hosted privatly. This is the the attempt to include this in the main arch linux arm repositary. Due to the number of iterations the number of changes is high:

1) 0001-add-lcd-cape-for-chiliboard.patch wasn't applying cleanly to the linux kernel -examining the code and it wasn't needed, so this is removed.
2) The dependence on linux firmware was examined, there are very few firmware patches relevant to arm, most of those mali related. None of these apply to the am335x target - so this dependence is removed. This saves a huge amount of space, and means that 1GB beagle bone black devices can again be support. Users will need to remove the linux firmware separately.
3) The old overlays (https://github.com/RobertCNelson/bb.org-overlays) was used, these are not consistent with modern kernel. As a result Robert 5 years ago moved onto
https://github.com/beagleboard/BeagleBoard-DeviceTrees/ - and in particular have tag versions for each kernel release. This updates the device tree overlays to Roberts tagged versions. This means that overlays will start working again on am335x devices, after several years of being broken on Arch Linux Arm.
4) In particular make sure we compile device tree with -@ so that relative paths can be followed in the device tree - this is needed for most overlays to load.
5) The firmware for am335x devices from Texas Instruments has been examined https://git.ti.com/cgit/processor-firmware. This is no longer maintained by TI on a heritage device, so the last released version is added to the package. (Note the TI github, does have one later tag which was not formally released. This may be used on a subsequent version). As TI no longer maintains this a snapshot has been taken and included in the package. 
6) The TI firmware blobs, and the am43x* blobs don't apply to am335x devices - so although included in the package, these are not including in the kernel build.

This has been tested both on a 1GB Beagle Bone Black (rev 5AC), and on a pocket beagle, and works without fault.


PhotoJim:
I've built this new 7.1.0 kernel for my BeagleBone Black and it seems to be working well so far. NOTE: it takes a long time (~44 hours!) to build on bare metal, so I'm going to figure out cross-compilation. It would be great if this could be merged.

PR #2209, opened 2026-07-26
linux-aarch64-rc: enabling modules for various rockchip and one allwinner peripherals i…

Popolon:
enabling modules for various rockchip and one allwinner peripherals including:

* RK3588: DP needed by second HDMI output port, MIPI DSI2, NPU (need to enable Rocket driver in mesa too for general usage), Rockchip Camera Interface (CIF), serial audio (SAI)
* RK3308: audio support
* RK3066: HDMI
* Allwinner H6: DMIC audio block

I use a RK3588, confirm than second HDMI now works with this change.


graysky2:
I will incorporate this when rc5 is released later today

PR #2212, opened 2026-08-08
core/linux-odroid-xu3 to 6.6.143-1

polyzen:


graysky2:
Tks for the PR. In the future please update the checksums for the config.

https://github.com/archlinuxarm/PKGBUILDs/commit/f749a5759acdb258fe59d37c783a8f7913703118

graysky2:
Also, no need to manually refresh the comments in the config unless something in it actually changes as a result of the upstream bump.

polyzen:
Sorry I keep forgetting to update its checksum lol. I just copy over the config from the build dir, but I can omit it when there are no real changes.

graysky2:
Not a big deal, but it's a good habit to get into, or better yet, in a script you write so you don't have to remember to.

PR #2219, opened 2026-09-03
core/linux-aarch64: enable linux tracing and event-tracing support

PolynomialDivision:
Enable standard Linux tracing and event-tracing support in the linux-aarch64 kernel configuration.


graysky2:
Implemented in 4018c80 - in the future, do not hand-edit config files and make a PR. You must have make completely refresh it. You skipped metric shit tons of config options.

PR #2224, opened 2026-09-14
alarm/rpi-eeprom: Add flashrom dependency to RPi4 and RPi5 packages

Relepega:
The `flashrom` package was not initially listed in the dependencies, while actually being required by the `rpi-eeprom-update` executable to flash an update.


graysky2:
Yeah, you sure about that? I've been flashing firmware with this package for both of those hardware SOCs without that forever.

Relepega:
yeah i did try every method possible (with rpi-eeprom-update only) for a couple of hours to flash the updated firmware on my pi5 but it failed every time. After installing the flashrom package it worked at 1st try. Maybe this isn't an issue on the pi4.

graysky2:
It works with or without it, but maybe doing it live with the new dependency is more straightforward without the reboot and temporary file cleanup.

